PR #7, opened 2018-09-07
Refactor parser

jwage:


lib/References/Doc.php
```diff
@@ -34,52 +34,23 @@ public function resolve(Environment $environment, string $data) : ?array
...
-
-                // recursively search all the children nodes for a match
-                foreach ($e['titles'] as $title) {
-                    if ($this->findEntryByText($title[1], $text)) {
-                        $entry = $e;
-                        break;
+            // try to resolve the link by looking in all the documents links
+            if ($entry === null) {
+                foreach ($metas->getAll() as $e) {
+                    foreach ($e['links'] as $name => $url) {
```
Ocramius: Holy nested loop batman!

Ocramius: Also: `$e` should be renamed

lib/MetaEntry.php
```diff
@@ -0,0 +1,123 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\RST;
+
+class MetaEntry
```
weaverryan: hell yes to this :D

PR #8, opened 2018-09-08
Throw an exception if inclusion fails

weaverryan:
```rst
.. include:: non-existent-file.rst
```

Current behavior: the file inclusion is silently skipped
New behavior: an exception is thrown

### DIFF
--- a/tests/FileIncluderTest.php
+++ b/tests/FileIncluderTest.php
@@ -1,0 +1,66 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\RST;
+
+use Doctrine\RST\Environment;
+use Doctrine\RST\FileIncluder;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+use RuntimeException;
+use function trim;
+
+class FileIncluderTest extends TestCase
+{
+    /** @var Environment|MockObject */
+    private $environment;
+
+    public function testInclude() : void
+    {
+        $this->environment->expects(self::once())
+            ->method('absoluteRelativePath')
+            ->with('include.rst')
+            ->willReturn(__DIR__ . '/files/include.rst');
+
+        $fileIncluder = new FileIncluder($this->environment, true, __DIR__ . '/files');
+
+        $contents = $fileIncluder->includeFiles('.. include:: include.rst');
+
+        self::assertSame('I was actually included', trim($contents));
+    }
+
+    public function testIncludeWithEmptyIncludeRoot() : void
+    {
+        $this->environment->expects(self::once())
+            ->method('absoluteRelativePath')
+            ->with('include.rst')
+            ->willReturn(__DIR__ . '/files/include.rst');
+
+        $fileIncluder = new FileIncluder($this->environment, true, '');
+
+        $contents = $fileIncluder->includeFiles('.. include:: include.rst');
+
+        self::assertSame('I was actually included', trim($contents));
+    }
+
+    public function testShouldThrowExceptionOnInvalidFileInclude() : void
+    {
+        $this->environment->expects(self::once())
+            ->method('absoluteRelativePath')
+            ->with('non-existent-file.rst')
+            ->willReturn('non-existent-file.rst');
+
+        $fileIncluder = new FileIncluder($this->environment, true, '');
+
+        $this->expectException(RuntimeException::class);
+        $this->expectExceptionMessage('Include ".. include:: non-existent-file.rst" does not exist or is not readable.');
+
+        $fileIncluder->includeFiles('.. include:: non-existent-file.rst');
+    }
+
+    protected function setUp() : void
+    {
+        $this->environment = $this->createMock(Environment::class);
+    }
+}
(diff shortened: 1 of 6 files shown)


lib/Parser.php
```diff
@@ -236,16 +236,12 @@ public function getDocument() : Document
         return $this->document;
     }
 
-    public function includeFileAllowed(string $path) : bool
+    private function includeFileAllowed(string $path) : bool
```
weaverryan: I'm under the impression that we're still fixing/improving this lib, so we aren't protecting BC. This method isn't used anywhere else in this project. I can also move this down to the bottom (below the public/protected functions) - I didn't want to make the diff unnecessarily big.

lib/Parser.php
```diff
@@ -236,16 +236,12 @@ public function getDocument() : Document
...
     {
         if (! $this->includeAllowed) {
             return false;
         }
 
-        if (! @is_readable($path)) {
-            return false;
-        }
```
weaverryan: Moved this so that it throws an exception. If an include is invalid, I think we should fail first. THEN we should check if the include is allowed.

lib/Parser.php
```diff
@@ -272,13 +268,13 @@ public function includeFiles(string $document) : string
...
+                if (!is_readable($path)) {
+                    throw new Exception(sprintf('Include path "%s" does not exist or is not readable.', $match[1]));
+                }
+
                 if ($this->includeFileAllowed($path)) {
                     $contents = file_get_contents($path);
 
-                    if ($contents === false) {
-                        throw new Exception(sprintf('Could not load file from path %s', $path));
-                    }
```
weaverryan: Because we're checking `is_readable()`, I'm not sure this is needed. But, it can be reverted if I'm wrong - not an important detail - trying to remove unnecessary stuff.

lib/Parser.php
Ocramius: Disagree with this change, because the method is called `includeFileAllowed`, which should not bother about whether it is called before or after other bits of code: it just responds ro the simple question on whether the path can be loaded.

lib/Parser.php
```diff
@@ -272,6 +268,10 @@ public function includeFiles(string $document) : string
             function ($match) {
                 $path = $this->environment->absoluteRelativePath($match[1]);
 
+                if (! is_readable($path)) {
```
Ocramius: What's the difference between throwing in the negative scenario of `includeFileAllowed` returning `false` and an explicit `is_readabl()` here?

Also, if inclusion is disabled, does `is_readable()` trigger stream wrappers?

tests/ParserTest.php
```diff
@@ -252,6 +252,15 @@ public function testInclusion() : void
         self::assertContains('I was actually included', $document->renderDocument());
     }
 
+    /**
+     * @expectedException Exception
```
Ocramius: Please declare expectations around exception within the method body

tests/ParserTest.php
```diff
@@ -261,6 +263,24 @@ public function testInclusion() : void
         self::assertContains('I was actually included', $document->renderDocument());
     }
 
+    public function testThrowExceptionOnInvalidFileInclude() : void
+    {
+        $this->expectException(RuntimeException::class);
```
Ocramius: Please move these right before the call that will throw

tests/FileIncluderTest.php
```diff
@@ -0,0 +1,101 @@
...
+        $fileIncluder = new FileIncluder($this->environment, $this->configuration, true, '');
+
+        $contents = $fileIncluder->includeFiles('.. include:: include.rst');
+
+        self::assertSame('I was actually included', trim($contents));
+    }
+
+    public function testShouldThrowExceptionOnInvalidFileInclude() : void
+    {
+        $this->expectException(RuntimeException::class);
```
Ocramius: Please move these right before the call that will throw

tests/ParserTest.php
```diff
@@ -261,6 +263,24 @@ public function testInclusion() : void
...
+
+        $parser      = new Parser();
+        $environment = $parser->getEnvironment();
+
+        $environment->getConfiguration()
+            ->throwExceptionOnInvalidFileInclude(true);
+
+        $data = file_get_contents(__DIR__ . '/files/inclusion-bad.rst');
+
+        assert($data !== false);
```
Ocramius: Please use the framework's test assertions: `self::assertInternalType('string', $data)` should work.

lib/Configuration.php
```diff
@@ -12,6 +12,9 @@ class Configuration
     /** @var bool */
     private $abortOnError = true;
 
+    /** @var bool */
+    private $throwExceptionOnInvalidFileInclude = false;
```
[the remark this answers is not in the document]
jwage: I will remove it. I was thinking originally that I might need it because the Doctrine docs might have bad includes sprinkled all over the place but I can just fix all of them before upgrading doctrine/rst-parser in doctrine/doctrine-website.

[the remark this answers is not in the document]
jwage: Fixed. And yes, lots of pre-existing weirdness. I plan on slowly working out the weirdness.

PR #10, opened 2018-09-18
Fix parsing tables

nikophil:
Hello, 
this PR fixes some problems with 
- differentiate some logic between simple table / pretty table (=grid tables)
- add thead for both
- prevent from providing malformed simple tables
- add some test case


weaverryan:
I think the scrutinizer failure is due to the complexity of an existing method. We could refactor that method now, but it would make this diff bigger.

tests/HTMLTest.php
```diff
@@ -74,45 +74,59 @@ public function testEmphasis() : void
...
 
+        self::assertSame(1, substr_count($document, '<thead>'));
+        self::assertSame(1, substr_count($document, '</thead>'));
         self::assertSame(2, substr_count($document, '<th>'));
         self::assertSame(2, substr_count($document, '</th>'));
         self::assertNotContains('==', $document);
     }
 
+    /**
+     * @expectedException \Exception
```
jwage: Can you set the expected exception with PHP code instead of annotations? e.g.

    $this->expectException(Exception::class);
    $this->expectExceptionMessage(...);

Also, lets set a more specific exception instead of `Exception`

lib/Nodes/TableNode.php
```diff
@@ -98,6 +117,10 @@ public function push(?array $parts, string $line) : bool
 
     public function finalize(Parser $parser) : void
     {
+        if (count($this->headers) === count($this->data)) {
+            throw new \Exception('Malformed table');
```
jwage: Throw a better more specific exception here. Probably `RuntimeException`

lib/Nodes/TableNode.php
```diff
@@ -47,6 +56,8 @@ public function getRows() : int
 
     /**
      * @param mixed[]|null $parts
+     *
+     * @throws \Exception
```
jwage: Add a use statement for the Exception class and remove the `\`

jwage:
It is ok for scrutinizer to fail. When you move around code, scrutinizer sees old problems as new ones and triggers a failure.

lib/Nodes/TableNode.php
```diff
@@ -47,6 +57,7 @@ public function getRows() : int
 
     /**
      * @param mixed[]|null $parts
+     *
```
jwage: Remove this extra new line.

lib/Nodes/TableNode.php
```diff
@@ -98,6 +116,10 @@ public function push(?array $parts, string $line) : bool
 
     public function finalize(Parser $parser) : void
     {
+        if (count($this->headers) === count($this->data)) {
+            throw new RuntimeException('Malformed table');
```
jwage: Thinking more about this error. When it happens, it will be hard to know where it is coming from? Thoughts @weaverryan?

nikophil: i totally agree
in the methods `finalize()` and `push()` in `\Doctrine\RST\Nodes\TableNode` we don't have access to the whole parsed table block...
wouldn't it be nice if we could have, node by node, the whole rst block, to perform some checks ?
and then, maybe print out the problematic block ?

jwage: In `TableNode` we have access to the `Parser` and `Environment` so we should be able to keep track of what the current file is and display that in the exception...I think. Can you investigate that approach?

nikophil: yep
maybe we could even print out the malformed table itself ? 
i don't know if we could access the line number in the current file, but it would be nice to have that as well

jwage: You should be able to print the malformed table. Should have all that data in the finalize() method. And yes, you are correct, at this point in the processing, we don't have the line number...since it is after all the lines have been parsed. I think just showing the file it is in and the contents of the malformed table will be good enough to merge this and in the future we can look at enhancements to try and detect the improperly formatted table earlier in the parsing so we can show file and line number.

PR #13, opened 2018-09-28
Allow Factory to be extended so nodes can be customized.

jwage:
Calling in the big guns (@Ocramius & @Majkl578) to possibly help figure out a better way to do this. Right now the Node objects that produce the HTML, LaTex, etc. from the parsed RST are hardcoded in this library. The node objects are created by the `Doctrine\RST\Factory` service. The change in this PR makes it so you can extend the `Kernel` and `Factory` so you can extend the node objects to customize the HTML that gets produced. I really want to find a better way to do this that doesn't involve inheritance and overriding public APIs but I am failing to see a better way currently. Thoughts?


lib/Kernel.php
```diff
@@ -9,28 +9,19 @@
...
+    protected $factory;
+
+    /** @var Directive[] */
+    protected $directives;
 
     /**
      * @param Directive[] $directives
      */
-    public function __construct(array $directives = [])
+    public function __construct(?Factory $factory = null, array $directives = [])
```
Ocramius: Swap these, or else you are inadvertent introducing a BC break

lib/Kernel.php
```diff
@@ -45,7 +36,15 @@ public function getFactory() : Factory
      */
     public function getDirectives() : array
     {
-        return $this->directives;
+        return array_merge([
```
Ocramius: This will re-instantiate all directives all the time: intentional?

lib/Factory.php
```diff
@@ -21,7 +21,7 @@ class Factory
     public const LATEX = 'LaTeX';
 
     /** @var string */
-    private $name;
+    protected $name;
```
Ocramius: I don't think this is the correct approach for extensibility, especially since this class is a lot of switching based on a constructor argument. Instead, adding an interface for the getters to allow consumers to provide their own  (arbitrarily complex) instantiation logic, as well as their own dependencies via constructor arguments, would be preferable.

Ocramius: Probably worth adding a note (at class level docblock) that this class is not meant to be extended

Ocramius: They can always wrap the original factory

jwage: Oh, right. Good call :+1:

lib/Kernel.php
jwage: I am going to fix this in this PR. They don't need to be newly created every time.

jwage:
@Ocramius @weaverryan @nikophil I think this gets us closer to something better. Could you all take a look and provide some feedback?

lib/Kernel.php
```diff
@@ -71,4 +80,19 @@ public function getFileExtension() : string
     {
         return 'txt';
     }
+
+    /**
+     * @return Doc[]
+     */
+    protected function createReferences() : array
```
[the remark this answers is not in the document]
nikophil: the parser is made in a way you have to extend this kernel if you want to add references / directives
ex: https://github.com/weaverryan/symfony-docs/blob/build-system/_build/src/HtmlKernel.php#L96

lib/NodeInstantiator.php
```diff
@@ -0,0 +1,49 @@
...
+        return $this->type;
+    }
+
+    /**
+     * @param mixed[] $arguments
+     */
+    public function create(array $arguments) : Node
+    {
+        /** @var Node $node */
+        $node = $this->reflectionClass->newInstanceArgs($arguments);
```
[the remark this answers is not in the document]
nikophil: i think because we don't know the number of arguments each type of node requires - this way we could pass it as an array

lib/Kernel.php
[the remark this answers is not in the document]
jwage: I changed this so end users can add custom references the same way you add custom directives, nodes, etc.

lib/NodeInstantiator.php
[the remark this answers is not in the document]
jwage: I think php allows you to do this other ways now without having to use call_user_func_array?

My thought process for using reflection was that it felt cleaner and I can refactor a little more to control the ReflectionClass instance in a unit test.

lib/NodeInstantiator.php
```diff
@@ -0,0 +1,49 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\RST;
+
+use Doctrine\RST\Nodes\Node;
+use InvalidArgumentException;
+use ReflectionClass;
+use function sprintf;
+
+class NodeInstantiator
```
nikophil: shouldn't it be wise to create an interface for this ? are we sure the `create()` function here will fit every needs ?

nikophil:
by the way, this makes things much cleaner and fits perfectly our needs, thanks :+1:

lib/NodeInstantiator.php
jwage: I wasn't intending for this to be extended, just used. You can use `DefaultNodeFactory` or implement your own instance of the `NodeFactory` interface. The`DefaultNodeFactory` just let's you configure and customize the class to use for each node type without having to implement your own `NodeFactory`.

jwage:
I'm going to work on adding tests for all of these changes now. Just wanted to get feedback on the direction before I finish it up.

lib/NodeInstantiator.php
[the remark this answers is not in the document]
jwage: I changed it to this if that is our preference instead of using reflection.

```
/** @var Node $node */
$node = new $this->className(... $arguments);
```

PR #16, opened 2018-09-30
add Environment accessors for currentFileName & currentDirectory

### LINKED ISSUE #15: Provide a Environment::getCurrentFileName()
Hello,

in order to have understandable error messages, we need to access to the envionment's current file, which does not have any accessor.

thank you.

nikophil:
fixes #15


jwage:
Would you mind trying to add a test for these methods?

PR #17, opened 2018-09-30
add $attributes to ResolvedReference

### LINKED ISSUE #14: Enhance ResolvedReference
Hello,

the class `Doctrine\RST\References\ResolvedReference` misses some way to add html attributes to the generated `<a>` 
Maybe we could add an `$htmlAttributes` associative array to this class, and then generate the expected markup in `\Doctrine\RST\Span::render()`?

thanks :)

nikophil:
fixes #14

### DIFF
--- a/lib/HTML/Span.php
+++ b/lib/HTML/Span.php
@@ -6,7 +6,11 @@ namespace Doctrine\RST\HTML;
 
 use Doctrine\RST\References\ResolvedReference;
 use Doctrine\RST\Span as Base;
+use function array_keys;
+use function array_map;
 use function htmlspecialchars;
+use function implode;
+use function sprintf;
 use function trim;
 
 class Span extends Base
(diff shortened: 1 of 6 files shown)


lib/HTML/Span.php
```diff
@@ -36,9 +37,17 @@ public function literal(string $text) : string
...
-    public function link(?string $url, string $title) : string
+    /**
+     * @param string[] $attributes
+     */
+    public function link(?string $url, string $title, array $attributes = []) : string
     {
-        return '<a href="' . htmlspecialchars((string) $url) . '">' . $title . '</a>';
+        $htmlAttributes = '';
+        foreach ($attributes as $attribute => $value) {
+            $htmlAttributes .= sprintf(' %s="%s"', $attribute, $value);
```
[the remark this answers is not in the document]
nikophil: do you want that i test if attribute names against a regex ? or maybe against a list of valid attributes ?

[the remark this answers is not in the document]
Ocramius: Any approach preventing invalid attribute names would be OK. An exception is ok if escaping is not possible.

lib/References/ResolvedReference.php
```diff
@@ -42,4 +52,21 @@ public function getTitles() : array
...
+
+    /**
+     * @return string[]
+     */
+    public function getAttributes() : array
+    {
+        return $this->attributes;
+    }
+
+    protected function validateAttributes() : void
```
Ocramius: `private`

Ocramius: Make the attributes a parameter, avoid setting them to property until after calling this method

tests/References/ResolvedReferenceTest.php
```diff
@@ -0,0 +1,141 @@
...
+
+        $this->assertSame('title', $resolvedReference->getTitle());
+        $this->assertSame('url', $resolvedReference->getUrl());
+        $this->assertSame(['title' => 'title'], $resolvedReference->getTitles());
+        $this->assertSame($attributes, $resolvedReference->getAttributes());
+    }
+
+    public function attributesValid() : \Generator
+    {
+        yield 'attributes #1' => [
```
Ocramius: Note: the `yield` approach is only useful when deferred data provider entries are required: favour a simple `array` when not needed.

nikophil:
hi @Ocramius 
i realized that we might also need to pass some attributes through the actual `Resolver`, then i made it and test the `Resolver`

PR #21, opened 2018-10-03
fix nested lists in tables

### LINKED ISSUE #20: Nested list inside table does not render
Hi,

there is a problem when nesting list inside table cell:

```rst
+--------+---------------+
| Cell 1 | Cell 2        |
+========+===============+
| Cell 5 | - List item 1 |
|        | - List item 2 |
|        | - List item 3 |
|        | - List item 4 |
+--------+---------------+
```

actual:
```html
<table class="table table-bordered">
    <thead>
    <tr>
        <th>Cell 1</th>
        <th>Cell 2</th>
    </tr>
    </thead>
    <tbody>
    <tr>
        <td>Cell 3</td>
        <td>- List item 1 - List item 2 - List item 3 - List item 4</td>
    </tr>
    </tbody>
</table>
```

expected:
```html
<table class="table table-bordered">
    <thead>
    <tr>
        <th>Cell 1</th>
        <th>Cell 2</th>
    </tr>
    </thead>
    <tbody>
    <tr>
        <td>Cell 3 </td>
        <td>
            <ul class="simple">
                <li>List item 1</li>
                <li>List item 2</li>
                <li>List item 3</li>
                <li>List item 4</li>
            </ul>
        </td>
    </tr>
    </tbody>
</table>
```

thank you :)

nikophil:
fixes #20


tests/HTMLTest.php
```diff
@@ -114,6 +114,20 @@ public function testHeaderTable() : void
...
     }
 
+    /**
+     * Testing HTML table with headers
+     */
+    public function testTableWithNestedList() : void
+    {
+        $document = $this->parseHTML('table-nested-list.rst');
+
+        self::assertSame(1, substr_count($document, '<ul>'));
```
Ocramius: Would it make sense to use Phpunit's DOM additions to check this?

nikophil: i don't know this plugin, but as you can see, there are a lot assertions about the DOM created by the parser in this file.
Maybe if it makes things easier, it would make sens

jwage: Maybe create an issue to improve and standardize how we are checking the dom in tests and we can make that change in another PR?

nikophil:
Getting a `ListNode` is not quite easy inside `TableNode`:
because parse is made line by line, there is now simple way to create a ListNode, and we need to fully parse (with `DocumentParser::parse()` the rst list). Then the idea is to create a way to parse fragments of rst, but while sharing the environment of the whole parser, that's why i created thie `Parser::parseFragment()` that does not replaces the main `$documentParser`

lib/Nodes/TableNode.php
```diff
@@ -104,7 +108,14 @@ public function finalize(Parser $parser) : void
             }
 
             foreach ($row as &$col) {
-                $col = $parser->createSpan($col);
+                $lines       = explode("\n", (string) $col);
+                $lineChecker = new LineChecker(new LineDataParser());
```
jwage: I would like to have this a dependency passed through instead of instantiating a new one here every time. But honestly, we sort of have the same issue already in DocumentParser where we create a new one there everytime. We can save this for another future round of refactoring.

nikophil: ok, you're right, i've passed it through the factory, then we're using the DocumentParser's instance

Issue #11, opened 2018-09-20
Node sent to SubDirective is always ParagraphNode


nikophil:
Hi,

the text provided for any SubDirective is always sent through a `ParagraphNode` 
This is quite annoying, when we want to add some class to the surrounding `<p>`, and even worse when we want to add some other markup - the only way to do it seems a little bit hacky... 

Example :
```rst
.. versionadded:: 4.1
    This option was introduced in Symfony 2.6 and replaces another option, which is available prior to 2.6.
```
Expected html would be:
```html
<div class="versionadded">
    <p>
        <span class="versionmodified">New in version 4.1: </span>
        This option was introduced in Symfony 2.6 and replaces another option, which is available prior to 2.6.
    </p>
</div>
```

All the surrounding div could be handled with a `WrapperNode`, except the `<span>`:
```php
    return new WrapperNode($document, '<div class="versionadded">', '</div>');
```

i would like to have your feedback about that ?

thanks a lot.

jwage:
This is how I have versionadded implemented for the Doctrine website https://github.com/doctrine/doctrine-website/blob/master/lib/RST/Directive/VersionAddedDirective.php

nikophil:
ok that's perfect, thanks!

PR #22, opened 2018-10-09
[WIP] Reference unresolved when inside some directives

nikophil:
Hi

@jwage here is the PR you've asked for

The problem is the following:
we'd like to have (for instance) `.. versionadded::` directive to output html like this:
```html
<div class="versionadded">
    <p>
        <span class="versionmodified">New in version 4.1: </span>
        blah blah blah
    </p>
</div>
```

This could not be achieve with a `WrapperNode`, because the ouput results with an unwanted `<p>`
Then, it have to be made with a `RawNode` and by casting the actual node as string:

```php
        return new RawNode(
            sprintf(
                '<div class="versionadded"><p><span class="versionmodified">New in version %s: </span>%s</p></div>',
                $data,
                strip_tags((string) $document, '<a><code>')
            )
        );
```

but then the document is rendered as html when casting, and the toctree is not constructed yet, so it does not know how to resolve potential `:ref:` inside the directive.

related to #11 

thanks :)


jwage:
This will need to be rebased and modified after https://github.com/doctrine/rst-parser/pull/13 was merged.

jwage:
I think the proper solution here is to actually create a new node instance that you can return. You could implement a VersionAddedNode and return that, and it wraps the other node and it doesn't call the other render() method until all the parsing is done. The CallableNode is a generic solution that end users can use when implementing custom directives. I updated your test directive to use it and your test passes now!

PR #23, opened 2018-10-16
tests case for links absolute / relative to files

nikophil:
Hello

@jwage here is the test cases about "relative" / "absolute" paths in toc tree that make fail the tests.

From Sphinx documentation: http://www.sphinx-doc.org/en/1.5.1/markup/toctree.html

Relative document names (not beginning with a slash) are relative to the document the directive occurs in, absolute names are relative to the source directory


weaverryan:
Ah, so the issue is that, currently, whenever we make a link, to find the file we're referring to, the rst-parser always assumes that we're pointing to the file using an absolute path, even if the path doesn't start with `/`. But really, it should look for the file relative if the path does not start with `/`.

lib/Builder/Copier.php
```diff
@@ -0,0 +1,77 @@
...
+            if (is_dir($source) && is_dir($destination)) {
+                $destination = dirname($destination);
+            }
+
+            if (is_dir($source)) {
+                $this->filesystem->mirror($source, $destination);
+            } else {
+                $this->filesystem->copy($source, $destination);
+            }
+        }
```
weaverryan: Should we clear the `$this->toCopy` when the method is finished? (same below for toMkdir)?

nikophil:
fixes https://github.com/weaverryan/symfony-docs/issues/7

PR #39, opened 2018-11-05
Convert standalone URIs and email addresses to links

ramsey:
According to the reStructuredText Markup Specification:

> A URI (absolute URI or standalone email address) within a text block is treated as a general external hyperlink with the URI itself as the link's text. ([ref])

This PR adds support for finding standalone URIs and email addresses and converting them to links.

[ref]: http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#standalone-hyperlinks

### DIFF
--- a/lib/Span.php
+++ b/lib/Span.php
@@ -159,11 +159,57 @@ abstract class Span extends Node
             return $id . $next;
         };
 
+        // Standalone hyperlink callback
+        $standaloneHyperlinkCallback = static function ($match, $scheme = '') use ($generator, &$tokens) {
+            $id  = $generator();
+            $url = $match[1];
+
+            $tokens[$id] = [
+                'type' => 'link',
+                'link' => $url,
+                'url' => $scheme . $url,
+            ];
+
+            return $id;
+        };
+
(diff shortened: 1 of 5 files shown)


ramsey:
Ugh. My filesystem (macOS) is set to be case-insensitive, so that `tests/html/` and `tests/HTML/` folder threw everything off on my end. I'll fix that.

ramsey:
Yep. No further changes from my end. Thanks!

jwage:
Just noticed this https://github.com/doctrine/rst-parser/blob/master/tests/HTMLTest.php#L144

ramsey:
Something odd is going on with those URLs. They're matched, but when replacing the contents with a link, the RST parser is doing something funky with the URLs. It looks like it's seeing `blah_` and trying to replace that with some randomly-generated token string. I think it's seeing that as a named link and assigning a parser link token to it, which causes it to break the URL.

jwage:
Ok, I can look at fixing that. I will create an issue for it.

jwage:
Logged here https://github.com/doctrine/rst-parser/issues/41

ramsey:
👍 

I had planned to create an issue for it, too. I wasn't going to let that sneak by. 😄

PR #42, opened 2018-11-05
More tests to show that the link regex is too eager.

### LINKED ISSUE #41: Fix bug requiring skipped test in HTMLTest
Need to fill in more details here but just logging it so we don't forget to fix this.

https://github.com/doctrine/rst-parser/blob/master/tests/HTMLTest.php#L144

jwage:
fixes #41

Adding more tests first to triage the issue. The link for replace links is too eager and matches any string with an underscore. So even before the support for standalone links, this would have happened:

```
This is a url: https://www.google.com/directory_
```

The `directory_` part would be seen as a link that needs to be replaced.

We need to figure out how to modify the regex so it isn't as eager.


tests/HTMLTest.php
```diff
@@ -727,7 +554,7 @@ public function testAnchors() : void
      */
     private function parse(string $file) : Document
     {
-        $directory   = __DIR__ . '/html/';
+        $directory   = __DIR__ . '/html-fixtures/';
```
ramsey: 💯

PR #58, opened 2018-11-17
Fix some edge case link

nikophil:


nikophil:
i've split this one into mutlliple PR

Issue #71, opened 2018-12-09
Include the file (and if possible line number) where an error occurs


weaverryan:
I know this could be tricky, at least the "line number" part. If you, for example, type a directive `.. foo::`, you will just get:

> Unknown directive: role  

I suppose we could store some "current context" somewhere so we can include it when the ErrorManager throws the errors?

wouterj:
@weaverryan didn't you fix this in your latest PR?

linawolf:
This was fixed by #73

PR #73, opened 2018-12-15
Throwing an unknown directive earlier and with more info

weaverryan:
Relates to #71 - this solves it in just one case... and it may be that we just need to add this contextual information piece-by-piece.

The if statement was moved because, before, it would find the directive, set it, then only explode LATER. This makes it explode as soon as it finds a directive that it doesn't know about (which allows us to print more contextual information).

### DIFF
--- a/lib/Parser/DocumentParser.php
+++ b/lib/Parser/DocumentParser.php
@@ -531,14 +531,6 @@ class DocumentParser
 
         $name = $this->directive->getName();
 
-        if (! isset($this->directives[$name])) {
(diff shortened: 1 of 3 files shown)


jwage:
Can we update the test to check the additional information about the file and line the unknown directive was found in?

weaverryan:
Updated! There's no existing test where the "Environment" actually contains the filename, unfortunately. So, we can add that - it's just a bit more work - not too familiar with the tests - but let me know :)

PR #74, opened 2018-12-16
adding test case for sections closing too far

weaverryan:
I noticed a minor issue with the `section` tags. Take the following situation:

```rst
h1 Foo
=====

h2 Bar
------

h3 Baz
~~~~~

h2 Bazzles
-----------
```

The `h2 Bazzles` should still be a child inside the `h1 Foo` section. But the HTML markup closes and puts the `h2` into a sibling div of the `h1`.

This includes a failing test case to show the behavior - but I'm not sure where this is controlled in the code.

Cheers!

### DIFF
--- a/lib/Parser/DocumentParser.php
+++ b/lib/Parser/DocumentParser.php
@@ -20,6 +20,7 @@ use Doctrine\RST\Parser\Directive as ParserDirective;
 use Exception;
 use Throwable;
 
+use function array_reverse;
 use function array_search;
 use function assert;
 use function chr;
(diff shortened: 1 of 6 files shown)


greg0ire:
@weaverryan I'm super late to the party but I think it is here: https://github.com/doctrine/rst-parser/blob/55253fc5088a7ec97123fc2a5258af41da7f2c02/lib/Parser/DocumentParser.php#L486-L496

greg0ire:
> we need to end all open sections 

No we don't! :stuck_out_tongue:

greg0ire:
I fixed the test but broke other tests… whack-a-mole!

greg0ire:
@weaverryan @wouterj please review carefully, it looks super suspicious to me that I had to edit these extra test files.

wouterj (approved):
Looks good!

Don't worry about having to edit other tests if you make a change: It's common in this package that the test files are copy-pasted from the output. E.g. the "titles" test is testing the `h1`, `h2`, etc. elements, the rest was just copy-pasted from the output and not looked at. In other words: tests may contain completely bad expectations

PR #76, opened 2018-12-23
Improve missing reference error

weaverryan:
Another error where, at the very least, I've added the current filename (if we have one) that's being parsed.

A few notes:

1) The message was duplicated... so that's handled
2) Eventually, we should also be adding the current line number of these errors - but I don't think that's really possible yet
3) The `getCurrentFileName()` returns the filename with the `.rst`... which makes it fine, but not perfect.

### DIFF
--- a/lib/Environment.php
+++ b/lib/Environment.php
@@ -143,7 +143,7 @@ class Environment
     public function resolve(string $section, string $data) : ?ResolvedReference
     {
         if (! isset($this->references[$section])) {
-            throw new InvalidArgumentException(sprintf('Unknown reference section %s', $section));
+            throw new InvalidArgumentException($this->getMissingReferenceSectionError($section));
         }
 
         $reference = $this->references[$section];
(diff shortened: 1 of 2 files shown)


jwage:
We need a test update for this change before we can merge it.

weaverryan:
Test added for 1 of the 2 situations. I stopped because I have a question: why does a missing reference in `resolve()` throw an exception, but in `found()` we go through the ErrorManager? What do each of these methods do? Are you against adding some phpdoc above the methods to describe their purpose?

jwage:
@weaverryan I think that should probably be changed to use the ErrorManager.

weaverryan:
That makes sense to me too - done!

lib/Environment.php
```diff
@@ -485,4 +487,13 @@ public static function slugify(string $text) : string
 
         return $text;
     }
+
+    private function addMissingReferenceSectionError(string $section) : void
+    {
+        $this->errorManager->error(sprintf(
+            'Unknown reference section "%s"%s',
+            $section,
+            $this->getCurrentFileName() !== '' ? sprintf(' in "%s" ', $this->getCurrentFileName()) : ''
```
jwage: I think we have this code in a few places now. Might be time to create a shared helper method? We can just log an issue for that to do in another PR.

PR #77, opened 2018-12-23
Adding support for references with ticks

weaverryan:
References are allowed to have extra "ticks" around them, which was already supported for normal links (see a few lines above in `LineDataParser`.

### DIFF
--- a/tests/Builder/input/index.rst
+++ b/tests/Builder/input/index.rst
@@ -19,6 +19,10 @@ Link to :ref:`the same doc reference <same_doc_reference>`
 
 .. _same_doc_reference:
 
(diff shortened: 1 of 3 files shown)


tests/Builder/input/index.rst
```diff
@@ -19,6 +19,10 @@ Link to :ref:`the same doc reference <same_doc_reference>`
 
 .. _same_doc_reference:
 
+Link to :ref:`the same doc reference with ticks <same_doc_reference_ticks>`
+
+.. _`same_doc_reference_ticks`:
```
jwage: This HTML needs to be asserted somewhere.

weaverryan: Yep, of course :)

weaverryan:
This should also be ready!

PR #79, opened 2018-12-24
Fixing anonymous reference over-matching

weaverryan:
We have some `__CLASS__` in the symfony docs, and they're being interpreted as anonymous links. I double-checked in Sphinx: we ARE parsing them correctly in this lib currently (e.g. the test is 100% correct), but we are *overmatching* - Sphinx correctly does NOT see `__CLASS__` as an anonymous link.

But, I'll be honest - I'm not sure how they're doing this - they seem to be smart enough to notice that there is not also a starting `__` at the beginning. I even tested `I love __GitHub__` to see if that would STILL match the anonymous link (after all, it still ends with `__`) but in Sphinx - it fails.

### DIFF
--- a/lib/Span/SpanProcessor.php
+++ b/lib/Span/SpanProcessor.php
@@ -13,6 +13,7 @@ use function preg_replace;
 use function preg_replace_callback;
 use function sha1;
 use function str_replace;
+use function substr;
 use function time;
 
 class SpanProcessor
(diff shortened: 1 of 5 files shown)


weaverryan:
I've added more test cases AND the fix. This is a blocker for the Symfony docs, because these are scattered across many docs in many versions - so nothing builds yet.

I did a lot more research into Sphinx and how it works. The given test cases reflect things correctly - you can add any punctuation you want a after the closing `__` and it is still seen as an anonymous link. But if you even add one `_` to the start, it's not a link anymore.

lib/Span/SpanProcessor.php
```diff
@@ -151,6 +156,11 @@ private function replaceLinks(string $span) : string
             /** @var string $link */
             $link = $match[3] ?: $match[5];
 
+            // a link starting with _ is not a link - return original string
+            if (substr($link, 0, 1) === '_') {
+                return $match[0];
+            }
```
weaverryan: This callback is also called for normal links, and so a weird link like this:

```
A link starting with an underscore: `_CLASS`_
```

WILL be affected by this change and will no longer be seen as a valid link. In Sphinx, this IS seen as a valid link, but I couldn't find any valid syntax at the bottom of the page to actually supply this URL - e.g. this doesn't work:

```rst
.. _`_CLASS`: https://symfony.com
```

So, I think this might not even be a supported case. We can't make this function ONLY affect anonymous links (I tried that) because after calling this callback for anonymous links, it is then called for normal links. That was causing the `__CLASS__` to THEN be seen incorrect as a normal link (e.g. for `__CLASS_`) that was broken.

weaverryan:
This one is also ready. It's more than just a failing test case now - it's the fix :)

PR #80, opened 2018-12-27
fix url enabled parameter

nikophil:
Hi,

i'm working on the `parse for pdf` feature for symfony docs, and i realized  something is wrong with this test, we're sending to the `baseUrlEnabledCallable` the path of the file we're actually treating, but, we want the path of the file on which we're making a link. This is the path we want to test against another path, to know if we're want to use the base url.

Example:
We would like to export all the `best practices` for pdf. Then, all links which are pointing outside from best practices need to be generated with the base url.

What do you think?

cheers :)


jwage:
I think this makes sense. @weaverryan is this what you all want/need? If so, just need tests updated/passing then we can merge.

nikophil:
tests are fixed

weaverryan:
> We would like to export all the best practices for pdf. Then, all links which are pointing outside from best practices need to be generated with the base url.

Yep, this is the correct situation we're looking for :)

PR #81, opened 2018-12-30
expose twig environment

nikophil:
Hello,

in order to create some nice and little more complex theme, we would need to register some twig filter / function.
Then we need the twig environment to be exposed.

This update implies to store the twig environment and not recreate an occurrence for each render. Hope it's not a problem.

thanks :)

### DIFF
--- a/lib/Templates/TwigAdapter.php
+++ b/lib/Templates/TwigAdapter.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\RST\Templates;
+
+use Doctrine\RST\Configuration;
+use Twig\Environment as TwigEnvironment;
+
+class TwigAdapter implements TemplateEngineAdapter
+{
+    /** @var Configuration */
+    private $configuration;
+
+    /** @var TwigEnvironment */
+    private $twigEnvironment;
+
+    public function __construct(Configuration $configuration)
+    {
+        $this->configuration = $configuration;
+    }
+
+    public function getTemplateEngine() : TwigEnvironment
+    {
+        if ($this->twigEnvironment === null) {
+            $this->twigEnvironment = TwigEnvironmentFactory::createTwigEnvironment($this->configuration);
+        }
+
+        return $this->twigEnvironment;
+    }
+
+    /**
+     * @param mixed[] $parameters
+     */
+    public function render(string $template, array $parameters = []) : string
+    {
+        return $this->getTemplateEngine()->render($template, $parameters);
+    }
+}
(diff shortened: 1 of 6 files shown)


Ocramius:
Why is a shared environment needed, specifically?

Also: tests strictly required, please :-)

nikophil:
do you want that we create some kind of `TemplateEngineAdapter` or something like that ?

Ocramius:
> i don't really understand your point... 

We're just trying to understand the underlying reasoning, before exposing library **internals** to userland. See also https://en.wikipedia.org/wiki/XY_problem, if X/Y wasn't clear.

nikophil:
indeed X/Y problem wasn't clear for me :)
ok so we need to register some extra twig function, because the theme we want to use for the preview layout comes with some assets, and then the path to these assets have to be relative to the current html file rendered. In order to perform that, a twig function seems to be the good solution - of course, the current path of the current file rendered must be exposed as well to the template.

jwage:
Would a better solution be to construct the `Twig_Environment` outside of `TwigTemplateRenderer` and inject it through the constructor? That way you would have access to pass it to other stuff before passing it to `TwigTemplateRenderer`? We def. need to do something better here, it may require other refactoring.

nikophil:
ok, i've tried something else.
Actually we cannot directly inject the `Twig_Environment` inside the `TwigTemplateRenderer` because we need the `Configuration::$customTemplateDir` and / or `Configuration::$theme` to be set before.
So we have to defer `TwigEnvironment` creation construct of the `Configuration` object.

I've also created a `TemplateEngineAdapter` in order to set the the bundle agnostic the template engine. This create a phpstan error because i can't set a typehint to `Configuration::getTemplateEngine()`

what do you think?

weaverryan:
@nikophil I think we should be precisely clear about what we need to accomplish that requires this PR - that will help everyone else determine if this is the best solution or something else.

I believe (but correct me if I'm wrong) that we currently just need these two abilities:
1) To add global variables (e.g. we will expose the current "version" of the docs that are being build)
2) A new function - I think this is needed to do some path manipulation. @nikophil can you give more details on this part? It may be important to knowing the best path.

Thanks!

jwage:
I think this one is somewhat related to https://github.com/doctrine/rst-parser/issues/83. Generally the themes & templating functionality needs work.

I created #87 to rethink how themes & templates work so that we can solve both #81 and #83 in a better way.

nikophil:
sorry, i've never seen the github notification about that...

> 2. A new function - I think this is needed to do some path manipulation. @nikophil can you give more details on this part? It may be important to knowing the best path.

actually a twig function is needed in order to compute the relative path between the current file rendered and the theme's assets folder.

lib/Templates/TwigAdapter.php
```diff
@@ -0,0 +1,35 @@
...
+namespace Doctrine\RST\Templates;
+
+use Doctrine\RST\Configuration;
+use Twig\Environment as TwigEnvironment;
+
+class TwigAdapter implements TemplateEngineAdapter
+{
+    /** @var Configuration */
+    private $configuration;
+    /** @var TwigEnvironment */
```
jwage: Please add a new line between these properties.

PR #82, opened 2018-12-30
failing test case for "raw urls"

nikophil:
Hi,
when the parser finds a trailing "raw url" in the rst files, it transforms it in an anchor node.

ex:
```rst
http://doctrine-project.org/
```

is tranformed to 
```html
<p><a href="http://doctrine-project.org/">http://doctrine-project.org/</a>
```

Problem is sometime in symfony doc, we use some urls that are not meant to be transformed.
Example:
`http://%env(HOST)%/project` that you can find here:
https://symfony.com/doc/current/configuration/external_parameters.html
(moreover it creates an anchor node in a code block, which seems pretty strange)

thanks :)

### DIFF
--- a/tests/Parser/ParserTest.php
+++ b/tests/Parser/ParserTest.php
@@ -389,6 +389,16 @@ class ParserTest extends TestCase
     }
 
     /**
+     * Assert that a raw url is not transformed into an anchor node
+     */
+    public function testRawUrl() : void
+    {
+        $document = $this->parse('url-raw.rst');
+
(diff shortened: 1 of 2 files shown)


jwage:
This is the intended behavior to convert links like this. This is a part of the RST specification and it was implemented by @ramsey in https://github.com/doctrine/rst-parser/pull/39. The main issue here I think is that we're converting to links in code blocks when we should not be? Can you update the test to reflect what is happening in Symfony?

PR #86, opened 2019-01-03
failing test case: directive nested in definition list

nikophil:
Hi,

a directive nested inside a definition list is not interpreted, and outputs not transformed:

```rst
term 1
    .. note::

        directive in definition list
```

is converted to:
```html
     <dl>
         <dt>term 1</dt>
         <dd>
            <p class="first">.. note:: </p>
            <p class="last">directive in definition list </p>
         </dd>
     </dl>
     </dl>
```

whereas sphinx works great with such case: https://symfony.com/doc/current/configuration/external_parameters.html#environment-variable-processors

here is a failing test case.

thanks

### DIFF
--- a/tests/Functional/tests/definition-list/definition-list.html
+++ b/tests/Functional/tests/definition-list/definition-list.html
@@ -73,3 +73,16 @@
         <dd>definition 1 definition 2 </dd>
     </dl>
 </div>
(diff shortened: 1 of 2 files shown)


wouterj:
This will be fixed by https://github.com/doctrine/rst-parser/pull/138 (also moved this commit there).

PR #89, opened 2019-01-05
create and dispatch OnLinkParsedEvent

nikophil:
Hello,

we would like to create a command which tests all external urls response code.
To achieve that goal, we need another hook point from which we can grab the link's url and the current file being parsed.

thanks.

### DIFF
--- a/lib/Event/OnLinkParsedEvent.php
+++ b/lib/Event/OnLinkParsedEvent.php
@@ -1,0 +1,43 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\RST\Event;
+
+use Doctrine\Common\EventArgs;
+
+final class OnLinkParsedEvent extends EventArgs
+{
+    public const ON_LINK_PARSED = 'onLinkParsed';
+
+    /** @var string */
+    private $url;
+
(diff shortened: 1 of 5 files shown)


lib/Event/OnLinkParsedEvent.php
```diff
@@ -0,0 +1,41 @@
...
+final class OnLinkParsedEvent extends EventArgs
+{
+    public const ON_LINK_PARSED = 'onLinkParsed';
+
+    /** @var string */
+    private $url;
+    /** @var string */
+    private $linkType;
+    /** @var string */
+    private $currentFileName;
```
weaverryan: I'm not positive about the Doctrine coding standards, but my guess is that the phpdoc above these is not needed/wanted

jwage: They are required so phpstan, etc knows what type the properties are so we can enforce type consistency.

jwage: We also need a new line inbetween each property.

nikophil: done ;)

PR #91, opened 2019-01-11
Re-adding caching of Metas & fixing how files are scanned and ref dependencies

weaverryan:
Hi!

I'm trying to re-add the caching of the metas files. It's fairly trivial, except that there are/were two problems:

1) In `Scanner`, the `$rst` variable is always missing its `.rst` extension. And so, the `file_exists($rst)` call always returned false.

2) Fixing that meant that a whole sub-system of "loading from an existing meta instead of re-parsing" was activated, and it's acting strangely:

```
filectime(): stat failed for /Users/weaverryan/Sites/os/rst-parser/tests/Builder/input//subdir/test-anchor.rst
```

You can see this in the new failed test: it's trying to get the `filectime` of some `test-anchor.rst` file. That is not a file, but actually an inline anchor from one of the source files (tests/Builder/input/subdir/index.rst):

```rst
This is a :ref:`test anchor <test-anchor>`

...

.. _test-anchor:

...
```

For some reason, this `test-anchor` is being seen as a dependency of the MetaEntry for this `subdir/index.rst` file. I don't know much about how the anchors are resolved and how this relates to dependencies, so I would love any guidance you may have.

Thanks!


jwage:
Currently I don't know. I remember seeing this weirdness where some things were seen as dependencies when they should not be and I have not tracked that down yet.

weaverryan:
I'm still checking into this - and have identified several related items.

weaverryan:
This is still WIP, but it's close and I'm now fixing a bunch of things.

Most important, the way that we were scanning for files was wrong and WAY too complex. I checked Sphinx: if i have a file `foo.rst` that is not referenced by ANY other files in ANY ways, it IS still rendered into HTML (but I get a warning that it's an "orphan" - not in any `toctree`). Our logic was super complex - we were following dependencies (toctree, doc & ref) in one file to find the next files.

I've now fixed and simplified that: `Scanner` simply finds all `.rst` files in a directory. Done. This made it much easier to figure out if a file needs to be parsed, or if its MetaEntry (and dependencies) are still fresh so it doesn't need to be.

lib/Meta/MetaEntry.php
```diff
@@ -115,7 +124,42 @@ public function getTocs() : array
      */
     public function getDepends() : array
     {
-        return $this->depends;
+        return array_values(array_unique($this->depends));
```
jwage: Is this needed here because of something upstream that should be fixed? Should the same value exist multiple times?

lib/Builder.php
```diff
@@ -198,4 +196,30 @@ private function render(string $directory, string $targetDirectory) : void
             new PostBuildRenderEvent($this, $directory, $targetDirectory)
         );
     }
+
+    private function loadCachedMetas(string $targetDirectory) : void
```
jwage: Can this be refactored out to a separate class that deals with the reading and writing of metas?

lib/Builder.php
```diff
@@ -198,4 +196,30 @@ private function render(string $directory, string $targetDirectory) : void
...
+    {
+        $metaCachePath = $this->getMetaCachePath($targetDirectory);
+        if (! file_exists($metaCachePath)) {
+            return;
+        }
+
+        $contents = file_get_contents($metaCachePath);
+
+        if ($contents === false) {
+            return;
```
jwage: I think you want to throw an exception here. If `file_exists` returns true and `file_get_contents` returns false, something is wrong.

jwage: Yes, an exception is what you want here and it will satisfy phpstan too.

jwage: It is technically possible for a systems permissions to return true for file exists but then false when you try to read so you want to let the end user know that with an exception instead of silently ignoring it.

weaverryan:
Feedback handled! :)

PR #92, opened 2019-01-11
Allowing directives to throw an exception and have a decent error

weaverryan:
If a directive needs to throw an exception for some reason, this gives that exception some context.


jwage:
This looks good. Can you rebase?

weaverryan:
Rebased! :)

PR #100, opened 2019-03-08
Adding support for colspan to tables

weaverryan:
I know the table parsing doesn't support many complex things. But, I've found a few "still-simple" use-cases that aren't parsing correctly yet.

Reference for this stuff: http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#grid-tables

Unless I've made a dumb mistake (possible, I've been staring at tables), the actual markup in these tests should be (in spirit) what Sphinx is rendering (minus things we know are different, like classes, `colgroup`, etc).


tests/Functional/FunctionalTest.php
```diff
@@ -115,7 +115,7 @@ public function getFunctionalTests() : array
                     ? 'renderDocument'
                     : 'render';
 
-                $tests[] = [$basename, $parser, $renderMethod, $format, $rst, trim($expected)];
+                $tests[$basename . '_' . $format] = [$basename, $parser, $renderMethod, $format, $rst, trim($expected)];
```
weaverryan: nice to be able to run `--filter=table-simple_html` for example

weaverryan:
This is still a WIP. I've done a ton of cleanup, and most of the work now will be in `TableNode`, to get the tests to pass when running:

```
./vendor/bin/phpunit tests/Functional/FunctionalTest.php --filter=table
```

lib/HTML/Renderers/TableNodeRenderer.php
```diff
@@ -27,45 +27,27 @@ public function __construct(TableNode $tableNode, TemplateRenderer $templateRend
...
+            unset($rows[$k]);
         }
 
         return $this->templateRenderer->render('table.html.twig', [
             'tableNode' => $this->tableNode,
-            'tableHeader' => $tableHeader,
-            'tableRows' => $tableRows,
+            'tableHeaderRows' => $tableHeaderRows,
+            'tableRows' => $rows,
         ]);
```
weaverryan: Because `$this->tableNode->getData();` now returns `TableRow[]` instead of some crazy array, I'm passing that straight to the template and dealing with it there.

lib/NodeFactory/DefaultNodeFactory.php
```diff
@@ -145,13 +145,10 @@ public function createListNode() : ListNode
         return $listNode;
     }
 
-    /**
-     * @param string[] $parts
-     */
-    public function createTableNode(array $parts, string $type, LineChecker $lineChecker) : TableNode
+    public function createTableNode(Parser\TableSeparatorLineConfig $separatorLineConfig, string $type, LineChecker $lineChecker) : TableNode
```
weaverryan: Changed this crazy `$parts` array (which was kind of a bunch of indexes identifying when each column started and stoped) into a new class.

lib/Nodes/Table/TableColumn.php
```diff
@@ -0,0 +1,72 @@
...
+
+namespace Doctrine\RST\Nodes\Table;
+
+use Doctrine\RST\Nodes\Node;
+use LogicException;
+use function strlen;
+use function trim;
+use function utf8_encode;
+
+class TableColumn
```
weaverryan: This, along with `TableRow` help `TableNode` work with meaningful objects, instead of a ton of arrays.

lib/Nodes/TableNode.php
```diff
@@ -4,177 +4,436 @@
...
+        if ($this->type === self::TYPE_SIMPLE) {
+            $this->compileSimpleTable();
         } else {
-            if ($this->type === self::TYPE_SIMPLE) {
-                $this->data[] = [];
+            $this->compilePrettyTable();
+        }
+    }
+
+    private function compileSimpleTable() : void
```
weaverryan: The old logic for turning the raw config/lines into a meaningful `$this->data` property were completely removed. I couldn't make any sense of them. I've replaced them with 2 entirely separate functions for the "simple table" and the "pretty/grid table". Each of these is super unfancy, because I want them to be readable and because algorithms are hard for me :).

tests/Functional/tests/simple-table-error/simple-table-error.html
```diff
@@ -1,8 +0,0 @@
-Exception: Exception
-Malformed table:
-
-First col | Second col | Third col
-Second row | Other col | Other col
-Third  row | Other col | Last col
-
-in file: "(unknown)"
```
weaverryan: File was renamed and made more interesting

lib/Parser/TableParser.php
```diff
@@ -41,53 +49,87 @@ public function parseTableLine(string $line) : ?array
...
             }
         }
 
-        $parts     = [];
-        $separator = false;
+        $parts = [];
+        /** @var int|null $currentPartStart */
+        $currentPartStart = null;
 
+        $i = 0;
```
weaverryan: Added this because phpstan thought that `$i` might not exist down on line 105 (after the for):

```php
$parts[] = [$currentPartStart, $i];
```

My impression was that `$i` would be initialized and set to 0, even if the `for` loop never looped: it's needed to make the comparison to see if the first loop is needed. So, I initialized it, unnecessarily (which scrutinizer is not happy about).

weaverryan:
CI is failing on nightly, which is unrelated.
Scrutinizer failures all look false.

This is ready to go :)

jwage:
Can you make all new classes final if they are not intended to be extended?

lib/NodeFactory/NodeFactory.php
```diff
@@ -65,10 +65,7 @@ public function createAnchorNode(?string $value = null) : AnchorNode;
 
     public function createListNode() : ListNode;
 
-    /**
-     * @param string[] $parts
-     */
-    public function createTableNode(array $parts, string $type, LineChecker $lineChecker) : TableNode;
+    public function createTableNode(Parser\TableSeparatorLineConfig $separatorLineConfig, string $type, LineChecker $lineChecker) : TableNode;
```
jwage: Can you use this class?

lib/Nodes/Table/TableColumn.php
jwage: Make this class final if you can.

lib/Nodes/Table/TableColumn.php
```diff
@@ -0,0 +1,72 @@
...
+{
+    /** @var string */
+    private $content;
```
jwage: Add new lines inbetween each class property.

lib/Nodes/TableNode.php
```diff
@@ -4,177 +4,435 @@
...
     protected $type;
 
     /** @var LineChecker */
     private $lineChecker;
 
-    /**
-     * @param string[] $parts
-     */
-    public function __construct(array $parts, string $type, LineChecker $lineChecker)
+    public function __construct(Parser\TableSeparatorLineConfig $separatorLineConfig, string $type, LineChecker $lineChecker)
```
jwage: Can you use this class?

lib/Nodes/TableNode.php
```diff
@@ -4,177 +4,435 @@
...
+                ->error(sprintf("%s\nin file %s\n\n%s", $this->errors[0], $parser->getFilename(), $tableAsString));
+
+            $this->data = [];
+
+            return;
+        }
+
+        foreach ($this->data as $i => $row) {
+            foreach ($row->getColumns() as $col) {
+                // TODO - not sure if this ever did anything
```
jwage: Can we do something about this? If not, lets remove it and add an issue for it to address later.

lib/Nodes/TableNode.php
```diff
@@ -4,177 +4,435 @@
...
+        foreach ($rows as $i => $row) {
+            if ($i > $finalHeadersRow) {
+                continue;
             }
+
+            $this->headers[$i] = true;
         }
+
+        $this->data = $rows;
+        // TODO - support rowspan
```
jwage: Can you create an issue in github for this and remove the TODO?

tests/ErrorManagerTest.php
```diff
@@ -18,8 +20,10 @@ public function testGetErrors() : void
             ->willReturn(false);
 
         $errorManager = new ErrorManager($configuration);
+        ob_start();
```
jwage: What is going on here?

jwage (changes requested):
Just a few changes and questions/comments. Looks good. Thanks!

lib/Nodes/TableNode.php
weaverryan: I removed it. I literally can't think of how that line was *ever* hit. So, it either never did anything, or it's so edge-case that I can't even figure out *what* issue I'd create. I think we roll forward without it, and see if any issues come up. Clearly we didn't have a test for the functionality it gave us, if any.

tests/ErrorManagerTest.php
weaverryan: When `isAbortOnError` is false, `ErrorManager` echos the errors... which literally means that content was being echo'ed during the test. The `ErrorManager` needs to be more awesome. This was just a way to silence the output in a test, but it *is* unrelated to this PR.

weaverryan:
Updates made!

jwage (approved):
:+1:

PR #101, opened 2019-03-08
Fixing invalid meta removal

weaverryan:
> The URL for :doc: links needs to be canonical to match when the dep is added. See `addDependency()` for matching logic.

> Also, if there are multiple invalid links, the dependency could be removed
multiple times - so we can't unset it once then check in removeDependency()
if it exists, because it would fail the second time.

Test is WIP. The bug basically only affects the cached meta files, which are used only for rebuilding faster with cache.

### DIFF
--- a/lib/Environment.php
+++ b/lib/Environment.php
@@ -157,8 +157,10 @@ class Environment
 
             if ($this->getMetaEntry() !== null) {
                 $this->getMetaEntry()->removeDependency(
-                    // use the unique, unresolved name
-                    $this->unresolvedDependencies[$data] ?? $data
+                    // for references, use the original, unresolved name
+                    // for doc links, use the canonical URL
+                    // see addDependency()
+                    $this->unresolvedDependencies[$data] ?? (string) $this->canonicalUrl($data)
                 );
             }
 
--- a/lib/Meta/MetaEntry.php
+++ b/lib/Meta/MetaEntry.php
@@ -153,8 +153,9 @@ class MetaEntry
     {
         $key = array_search($dependency, $this->depends, true);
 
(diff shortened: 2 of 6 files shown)


weaverryan:
This is ready. It's yet another thing that allows builds to successfully complete without aborting if you have `abortOnError` set to false.

lib/Environment.php
```diff
@@ -157,8 +157,10 @@ public function resolve(string $section, string $data) : ?ResolvedReference
 
             if ($this->getMetaEntry() !== null) {
                 $this->getMetaEntry()->removeDependency(
-                    // use the unique, unresolved name
-                    $this->unresolvedDependencies[$data] ?? $data
+                    // for references, use the original, unresolved name
+                    // for doc links, use the canonical URL
+                    // see addDependency()
+                    $this->unresolvedDependencies[$data] ?? (string) $this->canonicalUrl($data)
```
jwage: The return of canonicalUrl should be handled instead of casting to a string and silencing the problem. Exception?

lib/Meta/MetaEntry.php
```diff
@@ -153,8 +153,9 @@ public function removeDependency(string $dependency) : void
     {
         $key = array_search($dependency, $this->depends, true);
 
+        // it's possible a dependency is removed multiple times
         if ($key === false) {
-            throw new LogicException(sprintf('Could not find dependency "%s" in MetaEntry for "%s"', $dependency, $this->file));
+            return;
```
jwage: We should probably return something instead of void.

jwage:
Closing in favor of #104

PR #103, opened 2019-04-15
LogicException is surfacing when it should not with abortOnError = false

jwage:


lib/Environment.php
```diff
@@ -304,25 +307,27 @@ public function getLink(string $name, bool $relative = true) : string
...
-                $dependency
-            ));
-        }
-
         if ($requiresResolving) {
             // a hack to avoid collisions between resolved and unresolved dependencies
             $dependencyName                            = 'UNRESOLVED__' . $dependency;
             $this->unresolvedDependencies[$dependency] = $dependencyName;
+            // map the original dependency name to the one that will be stored
+            $this->originalDependencyNames[$dependency] = $dependencyName;
```
weaverryan: This is super tricky. This function is called first. And we basically store what dependencies this file has. But some are "docs"... which we change to be canonical right here (e.g. `foo` -> `subdir/foo`). Others are `:ref:` style, and we don't have enough info yet to resolve those. So, we do that resolving later.

In both cases, we're basically changing the input `$dependency` to "something else" before we store in `$this->dependencies`.

Then, LATER, the above `resolve()` method is called. If we have a problem resolving a dependency, we want to remove it from the `MetaEntry`. The problem is that the `MetaEntry` basically received `$this->dependencies`... which contains the *modified* dependency paths (thanks to this function). The `resolve()` function is passed the original dependency strings, which means we can't find them in `MetaEntry`. This map basically stores a map from the original dependency name to the new one, so that we can do a reverse lookup when we do `MetaEntry::removeDependency()`.

Admittedly, it's a mess. There are multiple things that need to happen in a specific, inconvenient order.

PR #106, opened 2019-05-03
Fixing bug where if a simple table row ended early, an error occurred

weaverryan:
Behavior verified in Sphinx.

### DIFF
--- a/lib/Nodes/Table/TableColumn.php
+++ b/lib/Nodes/Table/TableColumn.php
@@ -64,10 +64,19 @@ final class TableColumn
 
     public function render() : string
     {
-        return $this->getNode()->render();
+        $rendered = $this->getNode()->render();
+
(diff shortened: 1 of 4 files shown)


lib/Nodes/Table/TableColumn.php
```diff
@@ -64,10 +64,19 @@ public function setNode(Node $node) : void
 
     public function render() : string
     {
-        return $this->getNode()->render();
+        $rendered = $this->getNode()->render();
+
+        if ('' === $rendered && '\\' !== $this->content) {
+            $rendered = '&nbsp;';
```
weaverryan: This is the odd-ball. If a column is empty, then we put in an `&nbsp;`... unless the original content was `\`, in which case this means "please keep my column, but make it fully empty, no space"

jwage (approved):

PR #109, opened 2019-05-30
Implement 'initialHeaderLevel' config setting

monooso:
Implement support for a new 'initial header level' setting, which allows a user to control the 'level' at which document headings should start (`h1`, `h2`, etc.) Defaults to `1`.

The config setting is present in the docutils library, and is discussed briefly in [this StackOverflow comment](https://stackoverflow.com/a/33422946).

A common use-case for this enhancement is content-management systems which use separate fields for the page title (`h1`), and an RST-formatted body.


jwage:
Hi, thanks for the addition! Can you add docs for this new option here https://github.com/doctrine/rst-parser/blob/master/docs/en/configuration.rst

monooso:
@jwage Will do.

jwage (approved):

jwage:
Thanks!

PR #112, opened 2019-07-02
using file mtime instead of ctime for caching

weaverryan:
Fixes a bug if you, for example, rsync from one directory to another then build - the ctime might make those files look modified. We had this exact issue (we rsync from one dir to another before building): with ctime, all the source files looked modified *every* time.


jwage (approved):
💯

PR #113, opened 2019-07-04
Bumping twig version and removing a deprecation

weaverryan:
The spaceless tag was deprecated in 2.7. This upgrades the dependency and uses the new way of doing it so that anyone using this library can use Twig without the deprecation.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     "require": {
         "php": "^7.1",
         "symfony/filesystem": "^4.1",
-        "twig/twig": "^2.5",
+        "twig/twig": "^2.9.0",
         "doctrine/event-manager": "^1.0"
     },
     "require-dev": {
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -19,7 +19,7 @@
     "require": {
         "php": "^7.1",
         "symfony/filesystem": "^4.1",
-        "twig/twig": "^2.5",
+        "twig/twig": "^2.7.3",
```
weaverryan: There was some bug fix about the spaceless filter specifically on 2.7.3, so I thought we could just use that as the min.

jwage:
Looks like composer.json needs to be restricted a bit more?

weaverryan:
you're right. Updated - just the nightly build failing now

jwage (approved):

PR #115, opened 2019-12-09
Allow library to be included with Symfony 5 projects

mvriel:
The dependency on symfony/filesystem 4 makes it hard to include this
project in a Symfony 5 based project. As far as my tests have
determined, symfony/filesystem did not break BC between 4 and 5. As such
I have updated the version constraint to allow for both 4 and 5.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     ],
     "require": {
         "php": "^7.1",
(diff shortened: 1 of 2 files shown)


jwage:
Thanks!

PR #116, opened 2020-01-26
Allow parse and transform to be ran independently

mvriel:
For phpDocumentor we want to be able to build a TOC for the API
documentation and the output of one or more runs of this RST parser and
inject that into the Twig templates.

In the existing setup, this causes a chicken and egg problem since the
TOC is only available after parsing, but we want to add more to the TOC
based on the output of other modules.

The solution to this scenario is to enable executing the parse and
render phase individually. This will allow us to gather the TOC earlier
in the application, combine it with other TOC elements, and then later
render the output using this updated TOC.

### DIFF
--- a/lib/Builder.php
+++ b/lib/Builder.php
@@ -181,7 +181,7 @@ class Builder
         return $scanner->scan();
     }
 
-    public function parse(string $directory, string $targetDirectory) : void
+    public function parse(string $directory, string $targetDirectory): void
     {
         $this->ensureDirectoryExists($targetDirectory);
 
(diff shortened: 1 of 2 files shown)


mvriel:
Build failed due to a cache test (which did not fail locally?). I will look into it; I would still like feedback on my approach :)

greg0ire:
Should the newly public API be documented at https://github.com/doctrine/rst-parser/blob/0.1.x/docs/en/builder.rst ?

lib/Builder.php
```diff
@@ -220,8 +219,10 @@ private function parse(string $directory, string $targetDirectory, ParseQueue $p
         $parseQueueProcessor->process($parseQueue);
     }
 
-    private function render(string $directory, string $targetDirectory): void
+    public function render(string $directory, string $targetDirectory): void
```
SenseException, line 222: Render became public and should get its own tests.

SenseException (changes requested):
I don't know about the docs. It's current state shows how to build rst files in a simple way. Maybe there's place for a advanced section below that shows use cases where the other methods can be explained and why someone might want to use them.

PR #119, opened 2020-07-23
Fix link with a parenthesis before

jeremyFreeAgent:
Links like
```rst
Example (`Web Site`_), and Example (Link_)
```
were not rendered.

### DIFF
--- a/lib/Span/SpanProcessor.php
+++ b/lib/Span/SpanProcessor.php
@@ -202,14 +202,14 @@ class SpanProcessor
 
         // Replacing anonymous links
         $span = (string) preg_replace_callback(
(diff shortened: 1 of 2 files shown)


weaverryan (approved):

jeremyFreeAgent:
@weaverryan I thinks there is other same issues (like `"` for instance).

weaverryan:
@jeremyFreeAgent That would not surprise me at all. But this PR both fixes *a* bug and has a test case, so it's definitely good stand alone.

Would you prefer to fix a few other cases on this PR? Or merge it?

jeremyFreeAgent:
We can merge it as it may help users.

weaverryan:
Thanks Jeremy!

PR #120, opened 2020-07-30
Add options for CodeNode

jeremyFreeAgent:
This allow having options like `:name:`.

### DIFF
--- a/lib/Nodes/CodeNode.php
+++ b/lib/Nodes/CodeNode.php
@@ -15,6 +15,9 @@ class CodeNode extends Node
     /** @var string|null */
     protected $language = null;
 
+    /** @var string[] */
+    protected $options;
+
     /**
      * @param string[] $lines
      */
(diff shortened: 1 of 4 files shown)


lib/Nodes/CodeNode.php
```diff
@@ -15,6 +15,9 @@ class CodeNode extends Node
     /** @var string|null */
     protected $language = null;
 
+    /** @var string[] */
+    protected $options;
```
weaverryan, line 19: We should make it private since we're adding this as a new thing.

weaverryan (commented):
I like this, but what is unique about the CodeNode? What I mean is, i we're adding a `getOptions()` to `CodeNode`... why only that and not on *all* nodes? I'm not advocating that we do that, I'm just trying understand why (and if) CodeNode is a special case, or if it's not and we're only addressing one case, but not all of them in this PR.

Thanks!

jeremyFreeAgent:
Background info: I wanted to have some options as in Sphinx (https://www.sphinx-doc.org/en/1.5/markup/code.html).

weaverryan:
> Background info: I wanted to have some options as in Sphinx

Definitely. And I guess you then wanted access to those options in the template? And that's why you added `getOptions()` to `CodeNode`? Is that accurate?

I'm really not sure if having access to the options in the same way would be relevant for things *other* than the `CodeNode`. So I would be +1 for merging this with just the CodeNode addition.

jeremyFreeAgent:
> Definitely. And I guess you then wanted access to those options in the template? And that's why you added getOptions() to CodeNode? Is that accurate?

Yes exactly!

weaverryan:
Added last details in #123. Thanks!

PR #126, opened 2020-11-21
Migrate to ga

greg0ire:


tests/Functional/FunctionalTest.php
```diff
@@ -21,14 +21,22 @@
...
 
+use const LC_ALL;
+
 class FunctionalTest extends TestCase
 {
     private const RENDER_DOCUMENT_FILES = ['main-directive'];
 
+    protected function setUp(): void
+    {
+        setlocale(LC_ALL, 'en_US.utf8');
```
SenseException, line 37: Did you run into an error that is based on the locale in GitHub Action?

.github/workflows/static-analysis.yml
```diff
@@ -0,0 +1,44 @@
...
+        with:
+          path: "~/.composer/cache"
+          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Install dependencies with composer"
+        run: "composer install --no-interaction --no-progress --no-suggest"
+
+      - name: "Run a static analysis with phpstan/phpstan"
+        run: "vendor/bin/phpstan analyse --error-format=checkstyle | cs2pr"
```
stof: you should just use `vendor/bin/phpstan analyse`. It has native support for github actions now, outputting annotations when needed

tests/Functional/FunctionalTest.php
greg0ire: Yes, `grün` was normalized as `grn`

.github/workflows/static-analysis.yml
greg0ire: @stof it seems to work when there are no errors, but otherwise I get `Error: Start tag expected, '<' not found on line 2, column 2`. I recall experiencing this in the past, but I don't think it was with PHPStan.

greg0ire: I reverted to the previous solution, but you can still see the failing job at https://github.com/doctrine/rst-parser/runs/1443323655?check_suite_focus=true

stof: Well, you must remove the `| cs2pr` from the command (it is not included in my comment), as `cs2pr` is about converting checkstyle output to github actions annotations, and so makes no sense when the tool outputs annotations directly.

greg0ire: Oh okay, I feel dumb now :facepalm:

stof (approved):

tests/Functional/FunctionalTest.php
stof: maybe this indicates an issue in the tool, that should not rely on such locale...

PR #135, opened 2021-03-07
Fix permalinks of titles containing ticks

### LINKED ISSUE #45: Issue with <code> elements in headings
This file: https://github.com/sonata-project/SonataAdminBundle/blob/3.x/docs/reference/dashboard.rst has things like these:

```rst
The ``Admin`` list block
------------------------

Configuring the ``Admin`` list
------------------------------

Using the ``Admin`` service declaration
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

These are wrongly processed in the TOC, which shows the links like this:

image

The contents of the heading are correctly displayed, but the "weird alphanumeric string" leaks to the permalink too:

image

nikophil:
Hello, 

this fixes https://github.com/weaverryan/docs-builder/pull/45

The problem is related to what @weaverryan fixed here https://github.com/doctrine/rst-parser/pull/124

We actually need a way to access to the "real text" of the `SpanNode` from both `DocumentNode` and `TitleNode`, thus I've moved the logic into `SpanNode` with help of `SpanProcessor`. I've simplified a bit Ryan's `DocumentNode::getTextFromNode()` but no problem to take back his version if we think it's more efficient.

### DIFF
--- a/lib/Nodes/SpanNode.php
+++ b/lib/Nodes/SpanNode.php
@@ -17,6 +17,9 @@ class SpanNode extends Node
     /** @var string */
     protected $value;
 
+    /** @var string */
+    private $text;
+
     /** @var Environment */
     protected $environment;
 
@@ -43,6 +46,7 @@ class SpanNode extends Node
         $spanProcessor = new SpanProcessor($this->environment, $span);
 
         $this->value  = $spanProcessor->process();
+        $this->text   = $spanProcessor->getText($this->value);
         $this->tokens = $spanProcessor->getTokens();
     }
 
@@ -62,5 +66,10 @@ class SpanNode extends Node
     public function getEnvironment(): Environment
     {
         return $this->environment;
+    }
+
+    public function getText(): string
+    {
+        return $this->text;
     }
 }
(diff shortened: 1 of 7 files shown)


greg0ire:
:thinking: I don't think there are any forgotten PRs in 0.2.x, IMO what you want to target is 1.0.x

greg0ire:
Actually, you should rebase on 0.3.x.

greg0ire:
I think you might want to drop @weaverryan 's commits, as I probably had similar work to do on 0.3.x

lib/Nodes/SpanNode.php
```diff
@@ -17,6 +17,9 @@ class SpanNode extends Node
     /** @var string */
     protected $value;
 
+    /** @var string */
+    protected $text;
```
greg0ire: I don't think there is a reason for this to be protected.

```suggestion
    private $text;
```

Maybe we should make a lot of properties in this project private as well before releasing 1.0.0

nikophil: indeed you're right, fixed

javiereguiluz (approved):
Nice one Nicolas!

greg0ire (approved):

greg0ire:
Thanks @nikophil !

PR #138, opened 2021-03-14
Parse the contents of a definition list

wouterj:
This fixes the failing test case added by #86 

Previously, both the term and definition were parsed as span nodes (allowing only inline markup). However, the reStructuredText specs allows definitions to contain any body element ([ref](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#definition-lists)):

> ```
> +----------------------------+
> | term [ " : " classifier ]* |
> +--+-------------------------+--+
>    | definition                 |
>    | (body elements)+           |
>    +----------------------------+
> ```

I also added some PHPdoc to make it easier for future-me to understand how to parse sub contents (which will be needed e.g. to also fix lists: https://github.com/weaverryan/docs-builder/issues/38)

This change does change the contents of `DefinitionListTerm` and consequently how it needs to be rendered (see `lib/Templates/default/html/definition-list.html.twig`). Should we add a changelog to document these changes? And where can I find the Doctrine website theme to make sure we don't break that one?

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,0 +1,19 @@
+# Upgrade to 0.3
+
+## `DefinitionListTerm::$definitions` is a list of `Node`'s instead of `SpanNode`'s
+
+If you define a custom `definition-list.html.twig`, no longer wrap the value in
+a `<p>` element (the `.first` and `.last` classes are automatically added):
+
+```diff
+  <dl>
+      {% for definitionListTerm in definitionList.terms %}
+      <dt>{{ definitionListTerm.term.render()|raw }}</dt>
+      <dd>
+          {% for definition in definitionListTerm.definitions %}
+-             <p>{{ definition.render()|raw }}</p>
++             {{ definition.render()|raw }}
+          {% endfor %}
+      </dd>
+  </dl>
+```
--- a/lib/Parser.php
+++ b/lib/Parser.php
@@ -152,6 +152,9 @@ class Parser
         return $this->getNodeFactory()->createSpanNode($this, $span);
     }
 
+    /**
+     * Parses the given contents as a new file.
+     */
     public function parse(string $contents): DocumentNode
     {
         $this->getEnvironment()->reset();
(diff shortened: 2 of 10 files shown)


greg0ire:
> This change does change the contents of DefinitionListTerm and consequently how it needs to be rendered (see lib/Templates/default/html/definition-list.html.twig). Should we add a changelog to document these changes?

I was going to argue that we should wait for 1.0.0 before doing such things but I see that this library already has 11 dependents: https://packagist.org/packages/doctrine/rst-parser/dependents
On the other hand, I don't know of other Doctrine libraries where we keep a changelog. We seem to use `UPGRADE.md` when there are minor BC-breaks to document, let's try that instead? And since it's a "minor" BC-break (meaning we don't actually believe it will affect anybody in practice), targeting 0.3.x is correct.

> And where can I find the Doctrine website theme to make sure we don't break that one?


Weird. It does not appear in the dependent packages, but I think you might find what you are looking for in this repository: https://github.com/doctrine/doctrine-website

wouterj:
Thank you, I've added an `UPGRADE.md` following the syntax from `doctrine/dbal`.

> Weird. It does not appear in the dependent packages, but I think you might find what you are looking for in this repository: https://github.com/doctrine/doctrine-website

Ah, yes. I see it doesn't override the definition list template, so it's not affected (the same applies to `weaverryan/docs-builder`).

> I was going to argue that we should wait for 1.0.0 before doing such things but I see that this library already has 11 dependents: https://packagist.org/packages/doctrine/rst-parser/dependents

I'm not sure how this list of dependents is created, many of them don't have this package in either require or require-dev (e.g. `phpdocumentor/phpdocumentor`, `amelaye/biophp`, all `conduction/*` packages). At the same time, the current packages I'm aware of (doctrine's and Ryan's one) aren't listed as dependents :exploding_head:

greg0ire:
`amelaye/biophp` did seem to have a dependency at some point: https://packagist.org/packages/amelaye/biophp#v0.1.12 . I suppose that list is either not very fresh, or not only taking into account the latest versions of each package.

greg0ire (approved):
Love the comments, the upgrade note and the commit messages 👍

wouterj:
Oh, I think I messed up the target branch, or is it intended to release this as 0.3.1? (I wanted to apply the same fix for normal lists in another PR, guess that'll end up as 0.3.2 or 0.4.0 now :smile: )

Btw, #86 can be closed now.

greg0ire:
Well I didn't want to bother you with that so I made the mental gymnastics necessary to consider this a bugfix, I think we should use `0.4` when there is a big BC-break, but the difference between bugfixes and features will not really matter until we reach 1.0.0 IMO.

javiereguiluz:
Wouter, you did a fantastic job here. Thanks!

And thanks Grégoire for the review/merge and the maintenance work you do in this repository 🙏

PR #141, opened 2021-03-18
alphabetical globs, Directive refactoring & Fixing

weaverryan:
Hi!

This PR has gotten a big bigger!

It now has 5 things:

1) Fix alphabetical glob toctree (fixes https://github.com/weaverryan/docs-builder/issues/72)
2) Documented behavior
3) Discovery and Refactoring of Directive::wantCode()
4) Better errors & line numbers
5) Fixed directives allowing non-indented content (fixes https://github.com/weaverryan/docs-builder/issues/65)

**NOTE** This appears to not break the weaverryan/docs-builder test suite, which is also quite extensive :).

## 1) Fix alphabetical glob toctree

This fixes inconsistent toctree handling reported in https://github.com/weaverryan/docs-builder/issues/72

Sphinx, fortunately, is very clear that any "globbed" files in the toctree should be alphabetical: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#table-of-contents

> You can use “globbing” in toctree directives, by giving the glob flag option. All entries are then matched against the list of available documents, and matches are inserted into the list alphabetically

I have also rendered the test files in Sphinx directly to verify that we're doing the correct behavior.

## 2) Documented behavior

Along the, I documented a few things as I've discovered (through brute research) how they work ;).

## 3) Discovery and Refactoring of Directive::wantCode()

One BIG discovery is that the `Directive::wantCode()` causes those directives to be passed a `CodeNode`. However, this `wantCode()` method should probably not exist - I don't see the purpose of it. With a small fix to `DocumentParser` (made in this PR), directives can now safely return false from `wantCode()` and will be passed a `BlockNode` object (which is effectively identical to the `CodeNode` they were passed before). The `TocTree` directive was updated to show this new usage as well as `SubDirective`.

## 4) Better errors & line numbers

In a limited capacity, the library will now throw the original exception when there is an error as a nested exception.

AND, for the main line-by-line parsing, we now have line numbers in the errors!

It would probably be pretty easy to attach some (at least) rudimentary line number information (and filename info) to the Node objects, so that errors later could also have nicer info.

## 5) Fixed directives allowing non-indented content

Fixes: https://github.com/weaverryan/docs-builder/issues/65

Before this PR, this kind of stuff was allowed:

```rst
.. note::

I am note (get it?) indented
```

The parser would use the `I am note (get it?) indented` text as the "note" text. That is wrong - sphinx throws an error on this and the note is not rendered at all (and the `I am note (get it?) indented` is rendered as a normal paragraph).

This is now fixed. However, some directives - e.g. `class::` DO allow this craziness!

```
.. class:: align-center
 
I will be a "p" tag with an align-center class
```

To handle the behavior, there is a new `canApplyToNonBlockContent()` method on the `Directive` to control this. This, similar to (3) above, could cause BC breaks in any downstream libraries... but I don't know if there are any besides Symfony... and the behavior is currently, simply, wrong.

Thanks!


composer.json
```diff
@@ -30,7 +30,9 @@
         "phpstan/phpstan-phpunit": "^0.12",
         "phpstan/phpstan-strict-rules": "^0.12",
         "phpunit/phpunit": "^7.5 || ^8.0 || ^9.0",
-        "gajus/dindent": "^2.0.2"
+        "gajus/dindent": "^2.0.2",
+        "symfony/dom-crawler": "^5.2",
+        "symfony/css-selector": "^5.2"
```
weaverryan: I don't add dependencies (even dev dependencies) lightly. However, this made the tests in `BuilderTest::testToctreeGlobOrder` MUCH more precise and readable... and could probably be used in many other places in similar ways.

lib/Nodes/CodeNode.php
```diff
@@ -4,6 +4,27 @@
...
+ *          // I am the second line
+ *
+ * But a toctree is *also* considered a CodeNode
+ *
+ *      .. toctree::
+ *          :maxdepth: 1
+ *
+ *          file
+ *          file2
+ */
```
weaverryan, line 13: This was a big "ah ha!" moment for me. I had long thought it was a bug that `CodeNode` was used for some many things, like `toctree`. But apparently not - this node is just, perhaps, unfortunately named.

lib/Toc/ToctreeBuilder.php
```diff
@@ -42,13 +45,20 @@ public function buildToctreeFiles(
...
                     // if glob finds a file already explicitly defined
                     // don't duplicate it in the toctree again
                     if (in_array($globFile, $toctreeFiles, true)) {
                         continue;
                     }
 
+                    if ($globFile === $environment->absoluteUrl($environment->getCurrentFileName())) {
+                        // filter out the current file from being added as a glob
+                        continue;
+                    }
```
weaverryan, line 59: This is a behavior change in this library. It only affects glob toctrees:

```
.. toctree::
    :glob:

    *
```

Thanks to this change, this will no longer include the CURRENT file when matching the `*`. That is how Sphinx behaves.

lib/Toc/ToctreeBuilder.php
```diff
@@ -58,6 +68,10 @@ public function buildToctreeFiles(
             }
         }
 
+        if ((bool) ($options['reversed'] ?? false)) {
```
weaverryan, line 70: Apparently there is a `reversed` option that reverses the whole list. I verified our functionality matches Sphinx, and a test is included.

lib/Directives/Toctree.php
```diff
@@ -38,6 +41,8 @@ public function process(
             return;
         }
 
+        assert($node instanceof CodeNode, 'Toctree should be passed a CodeNode only');
```
weaverryan: This is enforced (in practice) by how the system works with `wantCode` below. If we were passed some other type of node, it would be unexpected. It allows us to also type-hint `CodeNode` instead of `Node` in `ToctreeBuilder`, which is good, because we can then get more information (like "options" such as `reversed`).

greg0ire:
If I understand correctly, this is a bugfix, so you should target 0.3.x

greg0ire:
> Along the, I documented a few things as I've discovered (through brute research) how they work ;).

Thanks to you and @wouterj for doing this :pray:

composer.json
greg0ire: Can we take this as an occasion to set `sort-packages` to true? It will make future diffs easier to understand.

lib/Directives/Toctree.php
wouterj: I believe this is a bug rather a feature, or a very wrong name of "CodeNode" :) (I think there is some wrong priority in how `DocumentParser` determines what the next block needs to be)

greg0ire: What should we do in the next major release? Make the type declaration line 35 stricter? Should we also add an `@param` annotation right now so that static analyzers can detect wrong calls?

lib/Directives/Toctree.php
```diff
@@ -53,6 +58,13 @@ public function process(
         $parser->getDocument()->addNode($tocNode);
     }
 
+    /**
+     * This directive includes the "code formatting" where the
+     * "value" of the block are indented as lines inside the directive.
+     *
+     * This will cause a CodeNode to passed to proces().
```
greg0ire:
```suggestion
     * This will cause a CodeNode to be passed to proces().
```

lib/Directives/Toctree.php
```diff
@@ -53,6 +58,13 @@ public function process(
         $parser->getDocument()->addNode($tocNode);
     }
 
+    /**
+     * This directive includes the "code formatting" where the
+     * "value" of the block are indented as lines inside the directive.
```
greg0ire:
```suggestion
     * "value" of the block is indented as lines inside the directive.
```

lib/Environment.php
```diff
@@ -386,6 +386,9 @@ public function setCurrentFileName(string $filename): void
         $this->currentFileName = $filename;
     }
 
+    /**
+     * Returns the currently-parsed filename, without the extension (e.g. "index" or "subdir/file")
```
greg0ire: So this file name is relative to something in particular? The working directory, or maybe the directory of some document being parsed?

lib/Nodes/CodeNode.php
greg0ire: :bulb: I see… what's the relationship between a CodeNode and a CodeBlock directive?

lib/Toc/GlobSearcher.php
```diff
@@ -15,6 +15,12 @@
 class GlobSearcher
 {
     /**
+     * Returns array of absolute URL strings.
+     *
+     * For example, if you were glob'inb from inside a file called
```
greg0ire:
```suggestion
     * For example, if you were glob'ing from inside a file called
```

lib/Directives/Toctree.php
wouterj: I'm afraid I don't have any time this week to dig deeper into this. The current parser has 3 very similar states/concepts: Directive, CodeNode and Block. I think we should do some cleanup there. I had a look last week and concluded this:

"CodeNode" is meant to be the indented text following `::` (called ["literal block" in reStructedText](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#literal-blocks)).
```
+------------------------------+
| paragraph                    |
| (ends with "::")             |
+------------------------------+
   +---------------------------+
   | indented literal block    |
   +---------------------------+
```

[Directives](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#directives) have a type, options and optionally a directive body (which, as the previous line ends with `::`, is considered a "CodeNode" by this library):
```
+-------+-------------------------------+
| ".. " | directive type "::" directive |
+-------+ block                         |
        |                               |
        +-------------------------------+
```

And the concept of "block" in the `DocumentParser` doesn't appear to have any related concept in reStructuredText.

lib/Nodes/CodeNode.php
weaverryan: GREAT question. 

The flow goes like this (it may be an over-simplification):

1) Document parser is actually parsing things line by line - https://github.com/doctrine/rst-parser/blob/0.3.x/lib/Parser/DocumentParser.php#L220. It first does some setup where if we're not currently already parsing something (e.g. like a table or code block) it tries to determine what this next thing will be. This part determines, for example, if the next thing might be a directive - https://github.com/doctrine/rst-parser/blob/0.3.x/lib/Parser/DocumentParser.php#L223-L289 - this step stores what the state is (table, list, directive) and, when applicable, information about what the directive is.

2) Once we are "in a block", that same `parseLine()` method (called now on the 2nd line of whatever we're in) looks at the STATE and creates different Node objects, like `CodeNode`, `ListNode`, `TableNode` etc. https://github.com/doctrine/rst-parser/blob/0.3.x/lib/Parser/DocumentParser.php#L470-L513

3a) After the Node is created in (2), IF we are NOT inside a directive, then that node is added to the document: https://github.com/doctrine/rst-parser/blob/0.3.x/lib/Parser/DocumentParser.php#L556

3b) After the Node is created in (2), IF we ARE inside a directive, it finds the correct `Directive` class and calls `process()` on it - https://github.com/doctrine/rst-parser/blob/0.3.x/lib/Parser/DocumentParser.php#L531-L537 - this is interesting because the job of this method is to ADD some Node to the Document. Basically, in the case of a directive, that Node is not automatically added to the Document, as is done for other nodes in step 3a. Instead, the `process()` method does it.

In step 3b - the `DirectiveProcces()` method - the directive ultimately should add a Node to the Document. But it can be ANY Node. In the case of the `CodeBlock` directive class (called specifically for `.. code-block::`, the original `CodeNode` created by `DocumentParser` is added to the Document - https://github.com/doctrine/rst-parser/blob/0.3.x/lib/Directives/CodeBlock.php#L55

But in the case of the `TocTree` directive class, it *extracts* information from the `CodeNode`, but ultimately creates its OWN `TocNode` and adds THAT - https://github.com/doctrine/rst-parser/blob/9f9887b282307c7cc8ff92a2c98aec39346696d9/lib/Directives/Toctree.php#L50-L53

I hope this helps! I'm trying to make public (here and in the code when I can) these... mysteries :). It's starting to make a lot more sense to me now.

lib/Directives/Toctree.php
weaverryan: > I believe this is a bug rather a feature, or a very wrong name of "CodeNode"

The more I look at it, the more I agree that this is a bug, more than a "bad name".

> "CodeNode" is meant to be the indented text following ::

Hmm, that may be true. It seems like:

A) Many directives are returning `true` from `wantsCode()` that should NOT. This is causing a `CodeNode` to be passed to them. `TocTree` is a perfect example. We could work to reverse those - return `false` from those - and instead allow them to work with some generic `Node` object that would be passed to them instead.

B) I'm... not entirely sure... why we need `wantsCode()` at all. Any directive should just be passed... some generic Node. Then it works its magic to create its specific node type (like `TocNode`). If you're right that the `CodeNode` is only meant to represent a literal block... that's just something that `DocumentParser` should handle... and it would not be seen as a directive.

FYI - I'm going to reverse some of these changes.

lib/Nodes/CodeNode.php
wouterj: Btw, this is somewhat loosely based on the original python-based parser (the directive process stuff however is completely specific for  this PHP-based parser).

The original parser is surprisingly well documented: https://sourceforge.net/p/docutils/code/HEAD/tree/trunk/docutils/docutils/parsers/rst/states.py#l43

lib/Parser/DocumentParser.php
```diff
@@ -479,11 +490,17 @@ private function flush(): void
...
+                    // This means we are in an indented area that is not a code block.
+                    // If we're NOT in a directive, then this must be a blockquote.
+                    // If we ARE in a directive, allow the directice to convert
+                    // the BlockNode into what it needs
+                    if ($this->directive === null) {
+                        $document = $this->parser->getSubParser()->parseLocal($node->getValue());
 
-                    $node = $this->nodeFactory->createQuoteNode($document);
+                        $node = $this->nodeFactory->createQuoteNode($document);
+                    }
```
weaverryan, line 525: This represents the key bug fix that allows directives to start returning false from `wantsCode()`. 

This is a edge case BC break (bit it's a fix): if a directive's `wantsCode()` currently returns `false`, then previously we would pass that directive a `QuoteNode` where the "value" is this `$document` variable. Now you would be passed a `BlockNode` where the value is just the string inside the code block.

But, the idea is that IF you want to actually "parse" the "content/value" of your directive as rst, then you should use a "sub parser". And, as far as I can tell, that's precisely the purpose of the `SubDirective` class: it takes in the raw string content/value, then parses it further. THAT is the proper location for this type of logic, I believe.

tests/Functional/tests/quote3/quote3.rst
```diff
@@ -1,3 +1,5 @@
 .. image:: /test.jpg
 
+Here is some text.
+
     This should be a quote, right?
```
weaverryan, line 5: I discovered (by testing in Sphinx) that the previous file was illegal: you an't have an image directive with content in it :)

lib/Nodes/CodeNode.php
greg0ire: > GREAT question

I'm very glad I asked it, and I have another one: how do you find the energy to write such qualitative answers? I was certainly not expecting something that stellar!

> The original parser is surprisingly well documented

Wow that's indeed great quality too!

lib/Parser/DocumentParser.php
```diff
@@ -479,11 +490,17 @@ private function flush(): void
                     /** @var string[] $lines */
                     $lines = $this->buffer->getLines();
 
-                    $blockNode = $this->nodeFactory->createBlockNode($lines);
+                    $node = $this->nodeFactory->createBlockNode($lines);
 
-                    $document = $this->parser->getSubParser()->parseLocal($blockNode->getValue());
+                    // This means we are in an indented area that is not a code block.
+                    // If we're NOT in a directive, then this must be a blockquote.
+                    // If we ARE in a directive, allow the directice to convert
```
greg0ire:
```suggestion
                    // If we ARE in a directive, allow the directive to convert
```

weaverryan:
Ok! This should be ready. I've updated the PR description... there are now several things mixed together. My apologies for the big PR, but I think we're "removing some evil" with this PR. Because of the size, I've changed back to 0.4 - there are some new features and there is a minor behavior change for some directives to fix the `wantCode()` bug.

About the `wantCode()` behavior of `Directive`, I have removed/reversed it on `SubDirective` and `TocTree` as examples. That's... probably already a really big PR... so I have not removed/revered it from everything else.

lib/Directives/SubDirective.php
```diff
@@ -30,10 +31,44 @@ final public function process(
...
+            /*
+             * A directive with no content (which is not allowed for sub-directives.
+             *
+             * For example:
+             *
+             *      .. note::
+             *
+             *      Now more normal text.
+             */
+            throw new RuntimeException('Content expected, none found.');
```
weaverryan: We actually had a test for this, but we were allowing it to create an empty note. Sphinx does not allow that: it issues a warning and renders no note.

lib/Directives/SubDirective.php
```diff
@@ -30,10 +31,44 @@ final public function process(
...
+         *      .. class:: special-list
+         *
+         *      - Test list item 1.
+         *      - Test list item 2.
+         *
+         * Notice the 2 list items are NOT indented. This is legal, and ultimately
+         * those two items would be parsed as a ListNode and THEN passed to
+         * ClassDirective (which extends SubDirective) for processing.
+         */
+        if ($node instanceof BlockNode) {
```
weaverryan, line 55: Switching `wantCode()` from `true` to `false` basically just mean that in the "normal" situation, the `CodeNode` became a `BlockNode`. So the logic here is the same as before... but I added a bunch of docs to describe why there is/was this if statement.

lib/Directives/Toctree.php
```diff
@@ -52,9 +52,4 @@ public function process(
 
         $parser->getDocument()->addNode($tocNode);
     }
-
-    public function wantCode(): bool
-    {
-        return true;
-    }
```
weaverryan, line 59: Most directives can change to `wantCode` false without any changes.

lib/ErrorManager.php
```diff
@@ -19,14 +20,23 @@ public function __construct(Configuration $configuration)
...
-    public function error(string $message): void
+    public function error(string $message, ?Throwable $throwable = null): void
     {
         $this->errors[] = $message;
 
         if ($this->configuration->isAbortOnError()) {
+            // throw the actual throwable if one is available
+            if ($throwable !== null) {
+                throw $throwable;
+            }
```
weaverryan: When the error comes from an actual exception, this is helpful because we can see the original stacktrace. I have only updated one place in the code so far to pass the exception.

lib/ErrorManager.php
```diff
@@ -19,14 +20,23 @@ public function __construct(Configuration $configuration)
...
+
             throw new Exception($message);
         }
 
+        if ($this->configuration->isSilentOnError()) {
+            return;
+        }
+
```
weaverryan: No more warnings while running the tests!

lib/Nodes/CodeNode.php
weaverryan: > I'm very gald I asked it, and I have another one: how do you find the energy to write such qualitative 
> answers? I was certainly not expecting something that stellar!

Lol, thank you ❤️ . I want some history... in case someone else wonders this in the future!

FYI - I have reworked this comment - I now believe that when CodeNode is passed to a non-code directive, it's basically an abuse (and probably all directives should return false from `wantCode()` and we should remove/deprecate that eventually).

tests/BuilderUrl/input/index.rst
```diff
@@ -8,4 +8,5 @@ Base URL
 .. toctree::
     :glob:
 
+    index
```
weaverryan, line 11: Because `*` no longer includes "this document" (which is a bug fix), I added `index` manually because the assertions were written to actually want that.

tests/BuilderWithErrors/BuilderWithErrorsTest.php
```diff
@@ -7,18 +7,20 @@
 use Doctrine\RST\Builder;
 use Doctrine\Tests\RST\BaseBuilderTest;
 
-class BuilderWithErrors extends BaseBuilderTest
+class BuilderWithErrorsTest extends BaseBuilderTest
```
weaverryan, line 13: This test lacked the `Test` suffix, and so was never run and was totally broken.

tests/BuilderWithErrors/BuilderWithErrorsTest.php
```diff
@@ -7,18 +7,20 @@
...
         $builder->getConfiguration()->abortOnError(false);
+        $builder->getConfiguration()->silentOnError(true);
     }
 
-    public function testMalformedTable(): void
+    public function testNoContentDirectiveError(): void
     {
-        $contents = $this->getFileContents($this->targetFile('index.html'));
-        self::assertStringContainsString('<table', $contents);
-        self::assertStringNotContainsString('<tr', $contents);
```
weaverryan: There was aa test asserting that a broken table doesn't render... except that since aa long time ago, we ARE able to render (at least mostly) broken tables :).

lib/HTML/Directives/Wrap.php
```diff
@@ -42,6 +43,11 @@ public function processSub(
         string $data,
         array $options
     ): ?Node {
+        // if there is a "Wrap" directive (e.g. a note::), blank content is unexpected
+        if ($document === null) {
+            throw new RuntimeException('Content expected, none found.');
+        }
```
weaverryan, line 49: In practice, from what I can see, this `Wrap` directive is used for the `note::` directive only. Most directives do no allow NO content. Note is one of those. This behavior is consistent with Sphinx (even the wording). Thanks to the error handling in `DocumentParser`, the user sees an error with the name of the file and approximate line number.

tests/Functional/tests/directive-title/directive-title.rst
```diff
@@ -1,3 +0,0 @@
-.. note::
-Some title!
-===========
```
weaverryan, line 3: This is TOTALLY illegal (even indenting it is). This is the wrong behavior. I fixed it, and deleted the test (this test would now issue an error about the note not having any content, which is covered already in another test.

lib/Directives/Directive.php
```diff
@@ -98,9 +98,41 @@ public function finalize(DocumentNode $document): void
...
+     *      .. class:: align-center
+     *
+     *      I will be a "p" tag with an align-center class
+     *
+     * If your directive allows the "class" directive functionality,
+     * return true from this function. The result is that your
+     * directive's process() method will be called for the next
+     * node after your directive (e.g. a ParagraphNode, ListNode, etc)
+     */
+    public function canApplyToNonBlockContent(): bool
```
greg0ire: This reads a bit like may or may not apply, is that really the case?

Otherwise let's maybe rename this to 
```suggestion
    public function appliesToNonBlockContent(): bool
```

Also, I had some doubts about whether "non-block content" was the right name (I now think it is).

To understand that, I dug a bit through the code until I stumbled upon this: https://github.com/doctrine/rst-parser/blob/7907706178f02198a423a907d5aa83dc2a356b70/lib/Parser/LineChecker.php#L57-L64

Now I'm confused… on line 63, `$line` is known to be `''`, so this should just `return true`? It looks like the entire method could be summed up to `return $line === '' || trim($line[0]) === '';`

lib/Directives/SubDirective.php
```diff
@@ -30,11 +30,34 @@ final public function process(
...
+            $subParser = $parser->getSubParser();
+            $document  = $subParser->parseLocal($node->getValue());
```
greg0ire: Let's get rid of the `$subParser` one-time variable, as it does not bring extra clarity over `getSubParser()`? 
```suggestion
            $document  = $parser->getSubParser()->parseLocal($node->getValue());
```

lib/Parser/DocumentParser.php
```diff
@@ -269,6 +284,18 @@ private function parseLine(string $line): bool
                         $separatorLineConfig = $this->tableParser->parseTableSeparatorLine($line);
 
                         if ($separatorLineConfig === null) {
+                            if ($this->getCurrentDirective() !== null && ! $this->getCurrentDirective()->canApplyToNonBlockContent()) {
+                                // If there is a directive set, it means we are the line *after* that directive
+                                // But the state is being set to NORMAL, which means we are a non-indented line.
+                                // Some special directive (like class) allow their content to be non-indented.
```
greg0ire:
```suggestion
                                // Some special directives (like class) allow their content to be non-indented.
```

lib/Parser/DocumentParser.php
```diff
@@ -618,6 +654,10 @@ private function initDirective(string $line): bool
         return true;
     }
 
+    /**
+     * Called on a NORMAL state line: it's used to determine of this
```
greg0ire:
```suggestion
     * Called on a NORMAL state line: it's used to determine if this
```

lib/Directives/Directive.php
weaverryan: Hey @greg0ire!

Good comments here :).

A) About the naming... I'm fine with your suggestion - I've updated it! :).

B) About `isBlockLine()` - you're totally right. this is a poorly written function. I decided not to touch it previously, but now that we're talking about it, I've updated it to your suggestion. This method has good unit tests... but I added one more test for the `''` use-case (which passed locally with the current code AND after my refactoring to your suggestion).

lib/Parser/DocumentParser.php
weaverryan: you're good at catching my many typos - thanks :)

weaverryan:
Hi again!

I've handled this last round of suggestions. I believe this is ready - I really appreciate the fast reviews!

Cheers!

tests/ToctreeBuilderTest.php
```diff
@@ -29,35 +31,43 @@ public function testBuildToctreeFiles(): void
...
-            ->with('test1')
-            ->willReturn('/test1');
-
         $this->globSearcher->expects(self::once())
             ->method('globSearch')
             ->with($environment, '*')
-            ->willReturn(['/test1', '/test2', '/test3']);
+            ->willReturn(['/test1', '/test2', '/test3', '/apple', '/current_filename']);
+
+        $environment->expects(self::any())
```
greg0ire: this kind of call doesn't actually do anything, let's remove it.
```suggestion
        $environment
```

greg0ire (commented):
Just one minor nitpick otherwise it looks great!

greg0ire:
@nikophil if you can give this a review too, that would be great!

tests/ToctreeBuilderTest.php
weaverryan: good tip!

greg0ire (approved):

lib/Directives/Directive.php
```diff
@@ -98,9 +98,41 @@ public function finalize(DocumentNode $document): void
 
     /**
      * Should the following block be passed as a CodeNode?
+     *
+     * You should probably return false from this. If you do,
+     * in most cases (unless you directive allows for some fancy
+     * syntax), you will receive a BlockNode object in processNode().
+     *
+     * See CodeNode for more details.
```
wouterj:
```suggestion
     * @see CodeNode
```

lib/Directives/SubDirective.php
```diff
@@ -30,11 +30,33 @@ final public function process(
...
+         * That is where it is legal to have a format like this:
+         *
+         *      .. class:: special-list
+         *
+         *      - Test list item 1.
+         *      - Test list item 2.
+         *
+         * Notice the 2 list items are NOT indented. This is legal, and ultimately
+         * those two items would be parsed as a ListNode and THEN passed to
+         * ClassDirective (which extends SubDirective) for processing.
```
wouterj, line 53: I don't think this is logic you introduced in this PR, but it is worth a second look (outside this PR) anyways (before releasing a new version).

Docutils implement this feature a bit less "magic":
* `.. class::` is a directive with or without content (like figure). If it doesn't have content, a "PendingNode" is inserted: https://sourceforge.net/p/docutils/code/HEAD/tree/trunk/docutils/docutils/parsers/rst/directives/misc.py#l373
* A transformer (classes that are run on the final DocumentNode, similair to e.g. Twig's node visitors) then visits all pending nodes and adds the class attribute to the next node: https://sourceforge.net/p/docutils/code/HEAD/tree/trunk/docutils/docutils/transforms/misc.py#l35

lib/Nodes/QuoteNode.php
```diff
@@ -4,6 +4,11 @@
 
 namespace Doctrine\RST\Nodes;
 
+/**
+ * Represents a "Block Quote"
+ *
+ * https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#block-quotes
```
wouterj:
```suggestion
 * @see https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#block-quotes
```

lib/Parser/DocumentParser.php
```diff
@@ -217,6 +223,11 @@ private function parseLines(string $document): void
         }
     }
 
+    /**
+     * Return true if this line has completed process.
+     *
+     * If false is returned, this function will be called again with the same line.
```
wouterj, line 229:
```suggestion
     * If false is returned, this function will be called again with the same line.
     * This is useful when you switched state and want to parse the line again
     * with the new state (e.g. when the end of a list is found, you want the line
     * to be parsed as "BEGIN" again).
```

lib/Parser/DocumentParser.php
```diff
@@ -479,11 +508,17 @@ private function flush(): void
                     /** @var string[] $lines */
                     $lines = $this->buffer->getLines();
 
-                    $blockNode = $this->nodeFactory->createBlockNode($lines);
+                    $node = $this->nodeFactory->createBlockNode($lines);
 
-                    $document = $this->parser->getSubParser()->parseLocal($blockNode->getValue());
+                    // This means we are in an indented area that is not a code block.
```
wouterj:
```suggestion
                    // This means we are in an indented area that is not a code block or definition list
```

Definitions lists are discovered by an indented line directly following a line with text. That is the only way it's distinguished from block quotes (which much be preceded by an empty line).

lib/Parser/DocumentParser.php
wouterj: I agree

lib/Parser/LineChecker.php
```diff
@@ -54,13 +54,17 @@ public function isListLine(string $line, bool $isCode): bool
...
+     * next lines.
+     */
     public function isBlockLine(string $line): bool
     {
-        if ($line !== '') {
-            return trim($line[0]) === '';
-        }
-
-        return trim($line) === '';
+        return $line === '' || trim($line[0]) === '';
```
wouterj, line 67: Is there a reason you removed the trime from `trim($line) === ''`? I would say that a line containing only whitespace should also be considered an empty line (trailing whitespaces do something occur).

lib/Parser/State.php
```diff
@@ -4,11 +4,23 @@
...
+/**
+ * "States" for DocumentParser as it parses line-by-line.
+ */
 class State
 {
-    public const BEGIN           = 'begin';
-    public const NORMAL          = 'normal';
-    public const DIRECTIVE       = 'directive';
+    /**
+     * There is currently no state: the next line will being a new state
```
wouterj: Is "will being" correct english? (sorry for asking a native english speaker, but it sounds like you maybe meant "will begin"?)

wouterj (commented):
Thank you Ryan. Lots of great stuff in here :)

lib/Parser/LineChecker.php
greg0ire: In that case, the second part of the `||` will detect it.

lib/ErrorManager.php
```diff
@@ -19,12 +20,16 @@ public function __construct(Configuration $configuration)
...
     {
         $this->errors[] = $message;
 
         if ($this->configuration->isAbortOnError()) {
-            throw new Exception($message);
+            throw new Exception($message, 0, $throwable);
+        }
+
+        if ($this->configuration->isSilentOnError()) {
+            return;
```
nikophil: don't we have some inconsistency here ? `abortOnError:true` && `silentOnError: true` may not behave as expected by the developer... maybe we could merge them both under the same parameter:
```
// \Doctrine\RST\Configuration

private const ERROR_STRATEGY_SILENT = 'SILENT';
private const ERROR_STRATEGY_THROW = 'THROW';

private $string $errorStrategy = self::ERROR_STRATEGY_SILENT;

public function isAbortOnError(): bool {
    return $this->errorStrategy === self::ERROR_STRATEGY_SILENT;
}

// ...
```

lib/Nodes/CodeNode.php
nikophil: shouldn't we at this point rename `CodeNode` class?

lib/Parser/DocumentParser.php
```diff
@@ -269,6 +284,18 @@ private function parseLine(string $line): bool
...
+                                // If there is a directive set, it means we are the line *after* that directive
+                                // But the state is being set to NORMAL, which means we are a non-indented line.
+                                // Some special directives (like class) allow their content to be non-indented.
+                                // But most do not, which means that our directive is now finished.
+                                // We flush so that the directive can be processed. It will be passed a
+                                // null node (We know because we are currently in a NEW state. If there
+                                // had been legitimately-indented content, that would have matched some
+                                // other state (e.g. BLOCK or CODE) and flushed when it finished.
+                                $this->flush();
+                            }
```
nikophil, line 300: very nice documentation :+1:

nikophil (approved):
very nice PR :+1:

lib/Directives/SubDirective.php
weaverryan: Ooooh, that `PendingNode` is indeed a much cleaner way to do that! The current implementation bothers me.

What I DID implement in this PR is a way to, at least, make this currently-odd behavior configurable - via the new `Directive::appliesToNonBlockContent()` method. This fixed the problem where this special/odd `.. class::` behavior was being allowed for ALL directives.

Later, if we can implement this in a more clean way, we could deprecate this new method. It's needed for now to fix the bad behavior.

lib/Parser/DocumentParser.php
weaverryan: I can see our knowledge about the library starting to intersect and help each other improve. Wonderful!

lib/Parser/State.php
weaverryan: it was definitely a typo! Being a native speaker makes me sloppy... tbh ;)

lib/Nodes/CodeNode.php
weaverryan: After some further discovery (you can see it on the updated `CodeNode` class phpdoc, I discovered that my assumption was wrong. `CodeNode` "should" === "code block". Currently, it is ALSO passed to ANY directive that returns `true` for `wantCode()`... but my thinking is that everyone should return `false` from `wantCode()` and that method should be removed. At that point, `CodeNode` would correctly become the node for code blocks only.

lib/ErrorManager.php
weaverryan: Hmm. I see what you're saying... but they do control two different mechanisms. For example, you could want to avoid throwing errors but you DO want the errors to be "echo'd". That doesn't fit into either the `SILENT` or `THROW` mode.

Buuuuut, you did make me realize that I have the logic a bit backwards in `ErrorManager`. I've just updated it so that if `silentOnError === false`, it WILL echo. Previously, I had the `echo` code AFTER the `throw`... so it would throw before it echo'ed.

That is a slight behavior change: if you WERE aborting on error previously, that will still happen, but now you will also see an echo first), but it makes these two mechanisms works correctly and independently.

weaverryan:
This should be ready now - Wouter and Niko were nice enough to review ❤️

greg0ire:
Great! Thanks @weaverryan ! I won't do a release because of @wouterj 's comment: https://github.com/doctrine/rst-parser/pull/141#discussion_r598263469 , but feel free to ask for one at any time!

weaverryan:
Thank you <3. We will coordinate - we're using the latest branch code for now anyways :)

PR #144, opened 2021-03-27
Close api

greg0ire:
There are a lot of protected variables that are not referenced in child classes, and most classes were final although they were never extended. `final` can of course be dropped in the future on classes that truly need it. Note that many classes are still not final because they are mocked in tests

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,14 @@
+# Upgrade to 0.5
+
+## Property visibility changed from protected to private
+
+Some directives and node classes had protected visibility for their properties.
+That has now been changed to private.
+
+## Final classes by default
+
+Many classes have been made final because they were never actually extended.
+
 # Upgrade to 0.4
 
 ## Refactored List Rendering
@@ -14,11 +25,6 @@ classes for HTML and LaTex, which receive the new `ListNode`.
 
 Removed `Doctrine\RST\Parser\ListLine` in favor of `Doctrine\RST\Parser\ListItem`
 and changed signature of `Doctrine\RST\Parser\LineChecker::isListLine()`.
-
-## Property visibility changed from protected to private
-
-Some directives and node classes had protected visibility for their properties.
-That has now been changed to private.
 
 # Upgrade to 0.3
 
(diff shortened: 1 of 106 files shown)


greg0ire:
@wouterj @nikophil please review

lib/HTML/Directives/ClassDirective.php
```diff
@@ -13,7 +13,7 @@
 use function array_map;
 use function explode;
 
-class ClassDirective extends SubDirective
+final class ClassDirective extends SubDirective
```
wouterj, line 16: We currently extend from this class in weaverryan/docs-builder

lib/HTML/Renderers/SpanNodeRenderer.php
```diff
@@ -13,7 +13,7 @@
 use function htmlspecialchars;
 use function trim;
 
-class SpanNodeRenderer extends BaseSpanNodeRenderer
+final class SpanNodeRenderer extends BaseSpanNodeRenderer
```
wouterj, line 16: Same here

wouterj (commented):
Only did a quick test run of this branch on weaverryan/docs-builder, to see if it breaks things. This may be used to reconsider keeping some things open for extension.

nikophil (approved):
:+1:

I checked into symfony's docs builder, I think it wouldn't be too much work to not extends `ClassDirective` nor `SpanNodeRenderer` and use composition instead

weaverryan:
@nikophil would you be available to put together a proof of concept PR on docs-builder that removes the extension of those classes?

nikophil:
Hi @weaverryan no problem, I may do that on this weekend

UPGRADE.md
```diff
@@ -15,6 +15,15 @@ classes for HTML and LaTex, which receive the new `ListNode`.
...
 
+## Property visibility changed from protected to private
+
+Some directives and node classes had protected visibility for their properties.
+That has now been changed to private.
+
+## Final classes by default
+
+Many classes have been made final because they were never actually extended.
+
```
wouterj: This probably has to target 0.5, now that 0.4 is released.

wouterj (approved):

UPGRADE.md
greg0ire: Oh right, absolutely!

PR #148, opened 2021-04-25
Adding an Error object

Nyholm:
The end goal of this PR is to be able to run the parser and output a well formatted error message. I would really like to be able to create good "github actions code comments" (or [commands](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-error-message)). That is why I need to be able to get the errors in a more structured way. 

This PR does not break BC.

### DIFF
--- a/lib/Environment.php
+++ b/lib/Environment.php
@@ -473,11 +473,17 @@ class Environment
         return $this->titleLetters;
     }
 
+    /**
+     * @deprecated use $this->getErrorManager()->addError() instead.
+     */
     public function addError(string $message, ?Throwable $throwable = null): void
     {
         $this->errorManager->error($message, $throwable);
     }
 
+    /**
+     * @deprecated use $this->getErrorManager()->addWarning() instead.
+     */
     public function addWarning(string $message): void
     {
         $this->errorManager->warning($message);
@@ -508,10 +514,9 @@ class Environment
 
     private function addMissingReferenceSectionError(string $section): void
     {
-        $this->errorManager->error(sprintf(
-            'Unknown reference section "%s"%s',
-            $section,
-            $this->getCurrentFileName() !== '' ? sprintf(' in "%s" ', $this->getCurrentFileName()) : ''
-        ));
+        $this->errorManager->addError(
+            sprintf('Unknown reference section "%s"', $section),
+            $this->getCurrentFileName()
+        );
     }
 }
--- a/lib/Environment.php
+++ b/lib/Environment.php
@@ -11,7 +11,6 @@ use Doctrine\RST\References\Reference;
 use Doctrine\RST\References\ResolvedReference;
 use Doctrine\RST\Templates\TemplateRenderer;
 use InvalidArgumentException;
-use Throwable;
 
 use function array_shift;
 use function dirname;
@@ -473,22 +472,6 @@ class Environment
         return $this->titleLetters;
     }
 
-    /**
-     * @deprecated use $this->getErrorManager()->addError() instead.
-     */
-    public function addError(string $message, ?Throwable $throwable = null): void
-    {
-        $this->errorManager->error($message, $throwable);
-    }
-
-    /**
-     * @deprecated use $this->getErrorManager()->addWarning() instead.
-     */
-    public function addWarning(string $message): void
-    {
-        $this->errorManager->warning($message);
-    }
-
     public static function slugify(string $text): string
     {
         // replace non letter or digits by -
@@ -514,7 +497,7 @@ class Environment
 
     private function addMissingReferenceSectionError(string $section): void
     {
-        $this->errorManager->addError(
+        $this->errorManager->error(
             sprintf('Unknown reference section "%s"', $section),
             $this->getCurrentFileName()
         );
--- a/lib/Error.php
+++ b/lib/Error.php
@@ -34,10 +34,10 @@ final class Error
     {
         $output = $this->message;
         if ($this->getFile() !== null) {
-            $output .= sprintf(' in "%s"', $this->file);
+            $output .= sprintf(' in file "%s"', $this->file);
 
             if ($this->line !== null) {
-                $output .= sprintf(' at line "%d"', $this->line);
+                $output .= sprintf(' at line %d', $this->line);
             }
         }
 
--- a/lib/Error.php
+++ b/lib/Error.php
@@ -1,0 +1,70 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\RST;
+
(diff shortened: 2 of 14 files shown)


Nyholm:
Oh.. my normal style of coding is to write a bunch of changes and then see if the CI is green. I guess Github's new "security feature" requires a maintainer's approval to run the CI on this PR.

lib/ErrorManager.php
```diff
@@ -20,9 +20,48 @@ public function __construct(Configuration $configuration)
         $this->configuration = $configuration;
     }
 
+    public function addError(Error $error): void
+    {
+        $this->errors[] = $error;
+
+        if (! $this->configuration->isSilentOnError()) {
+            echo '/!\\ ' . $error->__toString() . "\n";
+        }
```
wouterj: If we want to make an abstraction of the error handling in this library (which I agree with), I would say that we also move this code into an `StdErrorReporter` or the like (so that we can also e.g. add an `GitHubErrorReporter`).

lib/Error.php
```diff
@@ -0,0 +1,66 @@
...
+
+declare(strict_types=1);
+
+namespace Doctrine\RST;
+
+use Throwable;
+
+use function sprintf;
+
+class Error
```
greg0ire:
```suggestion
final class Error
```

lib/Error.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        $this->message   = $message;
+        $this->file      = $file;
+        $this->line      = $line;
+        $this->throwable = $throwable;
+    }
+
+    public function __toString(): string
+    {
+        $outout = $this->message;
```
greg0ire: Did you really mean to name this variable that way? :smile:

lib/ErrorManager.php
```diff
@@ -20,9 +20,48 @@ public function __construct(Configuration $configuration)
         $this->configuration = $configuration;
     }
 
+    public function addError(Error $error): void
+    {
+        $this->errors[] = $error;
+
+        if (! $this->configuration->isSilentOnError()) {
+            echo '/!\\ ' . $error->__toString() . "\n";
```
greg0ire:
```suggestion
            echo '⚠️ ' . $error->__toString() . "\n";
```

lib/Error.php
```diff
@@ -0,0 +1,66 @@
...
+
+    public function __construct(string $message, ?string $file = null, ?int $line = null, ?Throwable $throwable = null)
+    {
+        $this->message   = $message;
+        $this->file      = $file;
+        $this->line      = $line;
+        $this->throwable = $throwable;
+    }
+
+    public function __toString(): string
```
greg0ire: This does not seem to be used the magic way, so let's rename it?
```suggestion
    public function asString(): string
```

lib/ErrorManager.php
```diff
@@ -20,9 +20,48 @@ public function __construct(Configuration $configuration)
         $this->configuration = $configuration;
     }
 
+    public function addError(Error $error): void
```
greg0ire: Might be more handy to have a method that takes scalar arguments and builds an `Error` VO internally, instead of having `addError(new Error(…))` every time

lib/Error.php
Nyholm: No, My energy level was super low when I wrote this.

lib/ErrorManager.php
Nyholm: Sure

lib/ErrorManager.php
Nyholm: I pushed some changes without the abstraction. When we are adding a third format, then I think it starts to make sense with an abstraction. Right now it is just a simple if statement.

Nyholm:
Hm.. I need approval again..

greg0ire:
Consider making a PR that upgrades to `doctrine/coding-standard 9` or any other small thing, that way I can approve and merge it and you will be autonomous on this one ;)

Nyholm:
I am a doctrine org contributor already... 

Hm.. anyways: https://github.com/doctrine/rst-parser/pull/149

lib/ErrorManager.php
Nyholm: With this change, I wonder if I should use the existing `error`/`warning` method instead. The only downside is the argument order.. But I can live with `->error($message, $throwable, $file, $line)`

lib/Environment.php
```diff
@@ -508,10 +514,9 @@ public static function slugify(string $text): string
...
     private function addMissingReferenceSectionError(string $section): void
     {
-        $this->errorManager->error(sprintf(
-            'Unknown reference section "%s"%s',
-            $section,
-            $this->getCurrentFileName() !== '' ? sprintf(' in "%s" ', $this->getCurrentFileName()) : ''
-        ));
+        $this->errorManager->addError(
+            sprintf('Unknown reference section "%s"', $section),
+            $this->getCurrentFileName(),
```
greg0ire: This is not valid for old versions of PHP.
```suggestion
            $this->getCurrentFileName()
```

Nyholm: Thank you. I could not find that syntax error.

lib/ErrorManager.php
greg0ire: Or you could target 0.5.x and break BC :)

Nyholm: Which way will make it quicker for me to use this feature in my project?
 =)

greg0ire: Well we just released `0.4.0`, so it wouldn't make a lot of sense to release `0.5.0` with just that change. That being said, since `ErrorManager` is not final, adding arguments such as `$file` and `$line` would be a BC break for extending classes, so I'm not sure there is a choice either way.

greg0ire: I just merged https://github.com/doctrine/rst-parser/pull/144 that will make 0.5.0 less empty, and also make it less likely for such changes to become BC breaks.

Nyholm:
Could I ask you to do a 0.4.x -> 0.5.x merge?

Nyholm:
I've updated the PR to target 0.5.x. CI will be green after 0.4.x -> 0.5.x merge.

greg0ire:
I'm AFK until tonight, will try to!

greg0ire (approved):

Nyholm:
Thank you for the help. 

Im done with this PR now.

wouterj (approved):

greg0ire:
Thanks @Nyholm !

Nyholm:
Wohoo. Thank you for the reviews and merge!

PR #150, opened 2021-05-01
Update events.rst

voltel:
A few stylistic changes that IMHO might improve the document.
One particular change, is the use of the verb "invoke" and its past participle "invoked" when it comes to running the event handler. 
This is to follow collocations: dispatch an event, invoke an event handler, register/add an event listener, etc.

### DIFF
--- a/docs/en/events.rst
+++ b/docs/en/events.rst
@@ -63,40 +63,40 @@ method will be invoked every time a document is parsed:
 Available Events
 ----------------
 
-The events you can listen for are as follows:
+The events you can listen for and their respective handlers:
 
-- ``PreBuildScanEvent::PRE_BUILD_SCAN``
-    ``preBuildScan()`` method will be invoked 
-    before files are scanned, when using the builder.
+.. list-table:: 
+   :widths: 2 1 4
+   :header-rows: 1
+   :stub-columns: 1
 
-- ``PreBuildParseEvent::PRE_BUILD_PARSE`` 
-    ``preBuildParse()`` method will be invoked 
-    after files are scanned and before they are parsed, when using the builder.
-
-- ``PreBuildRenderEvent::PRE_BUILD_RENDER`` 
-    ``preBuildRender()`` method will be invoked 
-    after files are parsed and before they are rendered, when using the builder.
-
-- ``PostBuildRenderEvent::POST_BUILD_RENDER``
-    ``postBuildRender()`` method will be invoked 
-    after files are rendered, when using the builder.
-
-- ``PostNodeCreateEvent::POST_NODE_CREATE``
-    ``postNodeCreate()`` method will be invoked 
-    after a node is created.
-
-- ``PreParseDocumentEvent::PRE_PARSE_DOCUMENT`` 
-    ``preParseDocument()`` method will be invoked 
-    before a node is parsed.
-
-- ``PostParseDocumentEvent::POST_PARSE_DOCUMENT``
-    ``postParseDocument()`` method will be invoked 
-    after a node is parsed.
-
-- ``PreNodeRenderEvent::PRE_NODE_RENDER``
-    ``preNodeRender()`` method will be invoked 
-    before a node is rendered.
-
-- ``PostNodeRenderEvent::POST_NODE_RENDER``
-    ``postNodeRender()`` method will be invoked 
-    after a node is rendered.
+   *  - Event name 
+      - Invoked handler method
+      - Timing of event
+   *  - ``PreBuildScanEvent::PRE_BUILD_SCAN`` 
+      - ``preBuildScan()`` 
+      - before files are scanned, when using the builder.
+   *  - ``PreBuildParseEvent::PRE_BUILD_PARSE``
+      - ``preBuildParse()``
+      - after files are scanned and before they are parsed, when using the builder
+   *  - ``PreBuildRenderEvent::PRE_BUILD_RENDER``
+      - ``preBuildRender()``
+      - after files are parsed and before they are rendered, when using the builder
+   *  - ``PostBuildRenderEvent::POST_BUILD_RENDER``
+      - ``postBuildRender()`` 
+      - after files are rendered, when using the builder
+   *  - ``PostNodeCreateEvent::POST_NODE_CREATE``
+      - ``postNodeCreate()`` 
+      - after a node is created
+   *  - ``PreParseDocumentEvent::PRE_PARSE_DOCUMENT``
+      - ``preParseDocument()``
+      - before a node is parsed
+   *  - ``PostParseDocumentEvent::POST_PARSE_DOCUMENT``
+      - ``postParseDocument()`` 
+      - after a node is parsed
+   *  - ``PreNodeRenderEvent::PRE_NODE_RENDER``
+      - ``preNodeRender()``
+      - before a node is rendered
+   *  - ``PostNodeRenderEvent::POST_NODE_RENDER``
+      - ``postNodeRender()``
+      - after a node is rendered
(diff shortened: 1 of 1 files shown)


docs/en/events.rst
```diff
@@ -30,45 +30,73 @@ Add a new listener with the event manager:
...
+    ``postParseDocument()`` method will be invoked 
+    after a node is parsed.
+
+- ``PreNodeRenderEvent::PRE_NODE_RENDER``
+    ``preNodeRender()`` method will be invoked 
+    before a node is rendered.
+
+- ``PostNodeRenderEvent::POST_NODE_RENDER``
+    ``postNodeRender()`` method will be invoked 
+    after a node is rendered.
```
greg0ire: This looks like a list of 1-item definition lists (see https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#definition-lists) Should the ` - ` part be dropped?

voltel: I first used definition list (i.e. without the ``-`` part) but it didn't look nice, so here it is :)

greg0ire: > it didn't look nice,

Dumb question, but how did you check :sweat_smile: ?

voltel: I've recently watched a screencast by @weaverryan on [bundles](https://symfonycasts.com/screencast/symfony-bundle/routes-controllers) and created my first bundle - [VoltelExtraFoundryBundle](https://github.com/voltel/extra-foundry-bundle). During this endeavor, I learned two ways to inspect the results of my ``.rst`` documentation editing:

1. The [online editor](http://rst.ninjs.org/) which was mentioned [here](https://docutils.sourceforge.io/rst.html#id28).
2. I also installed Sphinx locally, as [described here](https://www.sphinx-doc.org/en/master/usage/quickstart.html), and used it to build documentation with their ``$ make html`` utility command. This is a time-consuming process because I needed to re-build the project after any change I wanted to see, but the building process was quite fast. It needs some "watch" functionality, similar to webpack, but I am not that good to set it up myself. 
   There is also a good entry on [Getting Started with Sphinx](https://docs.readthedocs.io/en/stable/intro/getting-started-with-sphinx.html) on "ReadTheDocs" site.

greg0ire: Very nice answer, thanks!

> it didn't look nice

Do you think there is any way we could make it nice? I suppose there is some concept of themes at some point, do you know what theme you are using (is it the same as on doctrine.org)? I'm asking because it feels quite wrong to have to write semantically incorrect RST (and generate semantically incorrect HTML) to make things look nice.

voltel: Thanks for nice feedback!

About themes: I have never created any Sphinx themes and, to be honest with you, I am not particularly good with ``css`` styling. 

After your question I did install the Doctrine theme from this [repository](https://github.com/doctrine/doctrine-sphinx-theme), and while the list of terms I previously committed for the events still looks fine, I re-discovered that the table would look also nice under this theme (it didn't look nice with the default ``alabaster`` theme in my Sphinx installation as the column ``widths`` configuration option I provided was ignored for some reason). 

On the side note, I don't thing there is any semantic _breach_ with the current layout as there is a list of events to describe, and for each event there is some information to be presented (i.e. a method name and time for this method invocation), which could be reflected with a definition list (unless you prefer to use definition lists only for glossaries). On the other hand, I might easily be wrong. 

Anyway, I included a file with a table (it has a file extension ``.txt`` because GitHub wouldn't allow we to attach it because of the "unsupported file extension", but contains an ``rst`` text). If you find the table layout to be more appropriate, I could make a new commit with the events presented in the table as in the file. 

[index.txt](https://github.com/doctrine/rst-parser/files/6412241/index.txt)

greg0ire: A Table is even better, please do that 👍 You can amend your commit, no need for a new one IMO.

> On the other hand, I might easily be wrong.

I might be wrong too, but when I see the following, I see either a wrapped list item, meaning there is one multiline item with nothing to separate the name from its "definition", or the indent is taken into account, in which case the list item is a definition list which just one item, which surely you will agree is not great. 

```
- ``PostParseDocumentEvent::POST_PARSE_DOCUMENT``
    ``postParseDocument()`` method will be invoked 
    after a node is parsed.
```

The table is a great idea, because we are not really dealing with definitions here.

voltel: > You can amend your commit, no need for a new one IMO.

I am relatively new to Git so I got lost with the steps I needed to do. I guess, I had to use ``git push --force`` from my local branch, but it was told on numerous ocasions not to use ``--force`` in public repos that I had no nerve to do it. Still have to learn a lot to feel safe with the Git. 😨 

Anyway, the table is in the last commit of ``patch-1`` branch.

greg0ire: Don't worry, I will squash merge :)

> I guess, I had to use git push --force from my local branch, but it was told on numerous occasions not to use --force in public repos that I had no nerve to do it.

Correct, `--force` was the way to go here. It's bad to use it on a public repo iff people are using the branch. For instance, if we force push `0.4.x`, that will be bad because it will confuse our contributors during rebase. In the case of a fork though, it's fine, because no one is supposed to be using your fork, so you can rewrite the history at will, it won't matter.

greg0ire (approved):

greg0ire:
Ping @wouterj :)

docs/en/events.rst
```diff
@@ -30,45 +30,73 @@ Add a new listener with the event manager:
...
-- ``PostParseDocumentEvent::POST_PARSE_DOCUMENT`` - Dispatches a method named ``postParseDocument()`` after a node is parsed.
-- ``PreNodeRenderEvent::PRE_NODE_RENDER`` - Dispatches a method named ``preNodeRender()`` before a node is rendered.
-- ``PostNodeRenderEvent::POST_NODE_RENDER`` - Dispatches a method named ``postNodeRender()`` after a node is rendered.
+The events you can listen for and their respective handlers:
+
+.. list-table:: 
+   :widths: 2 1 4
+   :header-rows: 1
+   :stub-columns: 1
+
```
wouterj: I don't think this directive is yet supported by the RstParser - at least a simple `grep list-table` didn't give any results.

It would be a nice directive to support though :)

voltel: I can replace this ``list-table`` with pseudo-graphics table made out of `+`, `-`, `|` and `=` characters. Shall I do it?

greg0ire: ping @wouterj

javiereguiluz: @voltel apart from `list-table`, RST defines two syntaxes for tables. One is very hard to maintain and requires drawing all table borders with `+`, `-`, etc.

I recommend you to use the other syntax, which is the first one shown here: https://docutils.sourceforge.io/docs/ref/rst/directives.html#tables

This is the one we use for example in Symfony Docs. See https://raw.githubusercontent.com/symfony/symfony-docs/5.4/console/verbosity.rst

linawolf: In the T3Docs we are offering 4 different table formats: 

https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/WritingReST/Reference/Content/Tables.html

For the complicated format there is a table editor: https://www.tablesgenerator.com/text_tables

linawolf: However with the data in Question I would probably not use a table but a format like confval (https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/WritingReST/Reference/Code/Confval.html)

linawolf (approved):

linawolf:
@greg0ire I rebased this on 0.6 and removed the table of the events to keep the current list. I think we can  merge it like this. What do you think?

greg0ire:
Sure!

PR #154, opened 2021-07-08
Fix parsing definition list term

wouterj:
This is catched by @javiereguiluz and reported in https://github.com/symfony-tools/docs-builder/pull/108 The classifier token should be " : " instead of ":".

~Tests are however failing: somehow, a space is lost. Twig renders with the space, but somewhere in the templating logic of this library, we remove the space.. I'm lost on where this happens.~

I also removed `spaceless`. It removes any whitespace between HTML tags, which is not expected for inline HTML tags (it removes any whitespace). In another PR, I'll remove the other usages of `spaceless` (in https://github.com/symfony-tools/docs-builder/pull/80#issuecomment-805878833 it was concluded that this never has the intended effect)

### DIFF
--- a/tests/Functional/tests/standalone-email-addresses/standalone-email-addresses.html
+++ b/tests/Functional/tests/standalone-email-addresses/standalone-email-addresses.html
@@ -80,7 +80,7 @@ Message-ID: &lt;<a href="mailto:1236@local.machine.example">1236@local.machine.e
 <blockquote>
     <hr />
     <dl>
-        <dt> Received <span class="classifier-delimiter">:</span><span class="classifier">from x.y.test</span></dt>
+        <dt>Received: from x.y.test</dt>
         <dd> by example.net
 via TCP
 with ESMTP
--- a/lib/Parser/LineDataParser.php
+++ b/lib/Parser/LineDataParser.php
@@ -11,6 +11,7 @@ use Doctrine\RST\Nodes\SpanNode;
 use Doctrine\RST\Parser;
 
 use function array_map;
+use function array_shift;
 use function count;
 use function explode;
 use function ltrim;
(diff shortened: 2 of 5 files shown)


greg0ire:
> Tests are however failing

:thinking: they look pretty good from where I sit… did you write this after seeing the build fail for 0f6d24d, and before it passed for e6205c9 ?

wouterj:
Yes, sorry, edited the message. `spaceless` was the issue here :)

tests/Functional/tests/standalone-email-addresses/standalone-email-addresses.html
```diff
@@ -80,7 +80,7 @@ <h1>
 <blockquote>
     <hr />
     <dl>
-        <dt> Received <span class="classifier-delimiter">:</span><span class="classifier">from x.y.test</span></dt>
+        <dt>Received: from x.y.test</dt>
```
greg0ire, line 83: Does this change really belong to this commit?

wouterj: Yes. As said before, this is the absolute crasiest test in this bundle: it's purpose is to test if an emailaddress in a sentence is transformed into a mailto link. It does this by pasting a raw mail message (see `standalone-email-address.rst`) with all headers as being "rst syntax". You can imagine that pasting some raw text as rst syntax leads to weird cases :)

This results in this line correctly being parsed as a definition list. The bug however was that we exploded on `":"` to find "definition classifiers". Instead, we should have used `" : "` (as you can see in the bug report from Javier). That means that this crazy raw email message also looses it's classifier.

greg0ire: I really mean "commit" here, not PR… your message talks about the relationship between this test and exploding with ` : `, yet test is touched in the commit that removes `{spaceless}`

greg0ire: Should it maybe have been touched in both commits?

greg0ire: Or should both commits be squashed maybe? It doesn't look right to have a commit with a failing build, then another commit that removes more than just spaces in this test.

wouterj: Oh, I see. Squashed the commits :heavy_check_mark:

javiereguiluz (approved):
Great! Wouter, thanks a lot for fixing this tricky bug.

greg0ire (approved):

greg0ire:
Thanks @wouterj !

Issue #156, opened 2021-08-13
Let's talk about joined effort


jaapio:
Thanks @stof for reaching out to the phpDocumentor team. I would like to talk to you guys for a moment about the doctrine/rst-parser. Which we integrated in phpDocumentor. More that a year ago we had contact with Jonathan Wage about the library. If I remember it correctly he forked the original library to a doctrine version. Back than there was nearly any maintenance on the lib. And from the doctrine team there was not much interest to support any additional features and improvements. That’s why we (@mvriel and me) have forked the doctrine version and started working on it.

With phpDocumentor we are trying to get a combination of API docs like we are rendering for years, with hand written manuals like doctrine and symfony are providing. In combination with the templates phpDocumentor provides and all static information we know about the code this should raise the level of documentation for PHP in general. That’s our goal.

Beside the missing support from the doctrine team which slowed us down in the additional features for phpDocumentor we wanted to do some large refactoring in the code to make it faster and fit better into the application workflow of phpDocumentor. So we have split the parsing and rendering in our version and some other improvements which allow us in the end to cache the parser results of each document, do some compile steps like we always do on cached items and then start the rendering. This includes also some major changes that are reducing the memory footprint of the library.

But now the there is some more movement on this project I think we can all benefit from the changes we are doing on both sides. And maybe bring it back into a single project which is maintained by all of us.

Please let me know if you are interested to join our forces and make some nice magic happen with the documentation of all the great projects out there.

greg0ire:
This sounds good to me! How do you propose we move forward with this? Are you going to send PRs to us, or would you rather have everyone work on a fork of `doctrine/rst-parser` in the `phpDocumentor` namespace, or some other namespace?

Consider joining `#rst-parser` at https://www.doctrine-project.org/slack

If you want to get merge permissions on this repository, that should be possible too.

wouterj:
Hi @jaapio! Thanks for this issue, I didn't know phpDocumentor maintained their own fork.

As you might know already, we've started using this to render some docs on symfony.com (e.g. twig and swiftmailer) and are planning to slowly integrate it in all doc build processes. One of the main reasons to use this repository instead of starting from scratch was because we would start with something that is (somewhat) battle-tested and has other communities that are involved in maintenance. I think joining efforts makes all the sense, as it would mean that in the end we get a better package, and with less efforts from each individual contributor.

I'll let you and the Doctrine team figure out what the best way to join efforts is :)

jaapio:
[...]

PR #158, opened 2021-09-08
Add more optional debugging features

javiereguiluz:
We're going to start using your great parser to build the entire Symfony Docs contents (from 2.0 to 6.0 version, so it's thousands of RST files and hundreds of thousands of lines of contents).

It's working great, but whenever there is an RST syntax problem in our docs, it's very hard to know exactly where or at least the file where it happened. That's why I had to add this in my local fork of `rst-parser` and I've created this PR in case you consider it useful for all.

Now I can run the doc build command with `-v`, `-vv` and `-vvv` to output increasingly more debugging info. Thanks!

### DIFF
--- a/lib/Builder/ParseQueueProcessor.php
+++ b/lib/Builder/ParseQueueProcessor.php
@@ -15,6 +15,8 @@ use function fwrite;
 use function getenv;
 use function sprintf;
 
+use const STDERR;
+
 class ParseQueueProcessor
 {
     /** @var Kernel */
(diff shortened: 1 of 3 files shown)


derrabus (approved):

greg0ire (commented):
Hi! Should these messages sent to stderr (which is really meant for such notifications despite its unfortunate name)? Also, since you plan on using this project heavily, you might be interested in #156

javiereguiluz:
Good idea Grégoire! I've just updated the code to write to `STDERR`. Thanks!

greg0ire:
Thanks @javiereguiluz !

PR #159, opened 2021-10-13
Add support for non-Latin scripts in titles

javiereguiluz:
We're parsing some contents in non-Latin scripts (`ar`, `ja`, `ru`, `uk`, `zh`) and it's failing because of the `slugify()` method of this project. See details: https://github.com/symfony-tools/docs-builder/pull/121

I propose to replace the custom `slugify()` code by the slugger provided by Symfony String. Please note that Symfony String requires PHP 7.2 and this project still required 7.1 as minimum, so I had to change that too. I hope it's OK.


fabpot (approved):

greg0ire:
> I hope it's OK.

IMO it's OK, this is not the kind of lib that should end up on production servers, but on pipelines of OSS projects.

.github/workflows/continuous-integration.yml
```diff
@@ -37,15 +36,15 @@ jobs:
 
       - name: "Install PHP with XDebug"
         uses: "shivammathur/setup-php@v2"
-        if: "${{ matrix.php-version == '7.1' }}"
+        if: "${{ matrix.php-version == '7.2' }}"
         with:
           php-version: "${{ matrix.php-version }}"
           coverage: "xdebug"
           ini-values: "zend.assertions=1"
```
greg0ire: You can drop this block entirely, since PHP 7.2 comes with PCOV
Alternatively, you can migrate to reusing a centralized workflow, like this: https://github.com/doctrine/common/pull/937/files

```yaml
    uses: "doctrine/.github/.github/workflows/continuous-integration.yml@1.1.0"
    with:
      php-versions: '["7.2", "7.3", "7.4", "8.0"]'
```

.github/workflows/continuous-integration.yml
```diff
@@ -37,15 +36,15 @@ jobs:
...
 
       - name: "Install PHP with PCOV"
         uses: "shivammathur/setup-php@v2"
-        if: "${{ matrix.php-version != '7.1' }}"
+        if: "${{ matrix.php-version != '7.2' }}"
```
greg0ire: This line should be dropped entirely.

tests/EnvironmentTest.php
```diff
@@ -86,4 +86,48 @@ public function getMissingSectionTests(): iterable
...
+    /**
+     * @dataProvider getTextsAndSlugs
+     */
+    public function testSlugify(string $text, string $expectedSlug): void
+    {
+        self::assertSame($expectedSlug, Environment::slugify($text));
+    }
+
+    /**
+     * @return mixed[]
```
greg0ire:
```suggestion
     * @return iterable<string, array{string, string}>
```

.github/workflows/continuous-integration.yml
```diff
@@ -37,15 +36,15 @@ jobs:
...
       - name: "Install PHP with XDebug"
         uses: "shivammathur/setup-php@v2"
-        if: "${{ matrix.php-version == '7.1' }}"
+        if: "${{ matrix.php-version == '7.2' }}"
         with:
           php-version: "${{ matrix.php-version }}"
           coverage: "xdebug"
           ini-values: "zend.assertions=1"
 
       - name: "Install PHP with PCOV"
```
greg0ire:
```suggestion
      - name: "Install PHP"
```

javiereguiluz:
Grégoire, thanks for your quick and detailed review. I did the requested changes.

greg0ire (approved):

weaverryan (approved):

greg0ire:
Thanks @javiereguiluz !

javiereguiluz:
Thanks for the merge! If I may ask, could your please release a new 0.4.2 version including this fix? Thanks a lot!

greg0ire:
Sure! Let me just migrate the other workflows to reusable workflows before that :slightly_smiling_face:

greg0ire:
See https://github.com/doctrine/rst-parser/pull/160

PR #160, opened 2021-10-14
Reuse workflows from upstream

greg0ire:
Not migrating the static analysis workflow since Psalm is not setup on
this repository yet.


wouterj (approved):
> Not migrating the static analysis workflow since Psalm is not setup on this repository yet.

IIRC, this repository uses PHPstan

greg0ire:
@wouterj on some repositories, we use both, that way we make sure we are understood by both tools.

greg0ire:
Oh no it does not work: :grimacing: https://github.com/doctrine/rst-parser/runs/3899626444?check_suite_focus=true

wouterj:
As you said on another PR, this package belongs to a category of packages which is not really part of any "application", but only the build process of a website/book. So I think it's fine to limit the static analysis as a way to improve the quality of the library (and not thrive to be compatible with any static analyzer a user my use).

wouterj:
> Oh no it does not work: :grimacing:

Seems like the secrets aren't available: https://github.com/doctrine/rst-parser/runs/3899626444?check_suite_focus=true#step:4:5

```
  with:
    command-name: laminas:automatic-releases:release
  env:
    GITHUB_TOKEN: ***
    SIGNING_SECRET_KEY: 
    GIT_AUTHOR_NAME: 
    GIT_AUTHOR_EMAIL: 
    SHELL_VERBOSITY: 3
```

greg0ire:
Ah I think they are supposed to be passed explicitly (which makes sense) https://docs.github.com/en/actions/learn-github-actions/reusing-workflows#passing-inputs-and-secrets-to-a-reusable-workflow

greg0ire:
Fix? https://github.com/doctrine/rst-parser/pull/161

PR #161, opened 2021-10-14
Explicitly pass secrets to release workflow

greg0ire:
I am not sure this is going to fix https://github.com/doctrine/rst-parser/runs/3899626444?check_suite_focus=true, but it would make sense not to pass secrets to another workflow by default.

### DIFF
--- a/.github/workflows/release-on-milestone-closed.yml
+++ b/.github/workflows/release-on-milestone-closed.yml
@@ -8,4 +8,10 @@ on:
 jobs:
   release:
     name: "Git tag, release & create merge-up PR"
-    uses: "doctrine/.github/.github/workflows/release-on-milestone-closed.yml@1.1.0"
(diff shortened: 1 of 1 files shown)


wouterj (approved):
Let's see if it works :)

greg0ire:
Ah actually, I think I will have to change the upstream workflow first: https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onworkflow_callsecrets

> If a caller workflow passes a secret that is not specified in the called workflow, this results in an error.

greg0ire:
https://github.com/doctrine/.github/pull/23

greg0ire:
It worked! https://github.com/doctrine/rst-parser/releases/tag/0.4.2

Now I have to fix all the repos I changed :weary:

PR #163, opened 2021-10-15
Relax the requirement of symfony/translation-contracts dependency

javiereguiluz:
I introduced this dependency myself because it's required by `symfony/string` ... but I made it too restrictive. Without this change we can't use the symfony-tools/doc-builder in a Symfony app because it uses Symfony Translation ^5.3, which in turn requires translation-contracts ^2.0.

I'm sorry for all this noise 😞


greg0ire (approved):

greg0ire:
No worries! Released as https://github.com/doctrine/rst-parser/releases/tag/0.4.3

PR #165, opened 2021-10-17
Fixing bug where indented table header was seen as a dl

weaverryan:
Hi!

We found this small bug from the Symfony documentation. If a simple table's header started with a space:

```rst
======  ===== =====
 Col A  Col B Col C
======  ===== =====
Col X   Col Y Col Z
------  ----- -----
Col U   Col J Col K
======  ===== =====
```

Then the entire structure was seen as a definition list, instead of a table. The fix is pretty logical: we check FIRST if the top line (e.g. `===== ===== ====`) is a valid "table separator". And only if it is NOT, we then check to see if the next line (e.g. ` 
COL A COL B COL C`) is indented to trigger a definition list.

Cheers!


greg0ire (approved):

greg0ire:
Thanks @weaverryan !

lib/Parser/DocumentParser.php
```diff
@@ -284,15 +284,17 @@ private function parseLine(string $line): bool
                         $this->buffer = new Buffer();
                         $this->flush();
                         $this->initDirective($line);
-                    } elseif ($this->lineChecker->isIndented($this->lines->getNextLine())) {
-                        $this->setState(State::DEFINITION_LIST);
-                        $this->buffer->push($line);
-
-                        return true;
                     } else {
```
wouterj, line 287: Wouldn't it be more logical if we slightly reorganize these nested if statements?

```php
} elseif ($separatorLineConfig = $this->tableParser->parseTableSeparatorLine($line)) {
    $this->setState(State::TABLE);

    $tableNode = $this->nodeFactory->createTableNode(
    // ... (from line 313)
} elseif ($this->lineChecker->isIndented($this->lines->getNextLine())) {
    $this->setState(State::DEFINITION_LIST);

    // ... (from line 291)
} elseif ($this->getCurrentDirective() !== null && ! $this->getCurrentDirective()->appliesToNonBlockContent()) {
    // ... (from line 298)
} else {
    $this->setState(State::NORMAL);

    return false;
}
```

weaverryan: Almost definitely :). Please propose if if you have a vision - I'm a big fan of incrementally tweaking this library simply in order to make the code more understandable

PR #166, opened 2021-10-20
Failing test case for references

wouterj:
This adds a failing test for a bug we are experiencing on symfony.com: In articles inside a directory, the implicit references (`` `Some Headline within This Page`_ ``) generate a wrong relative URL (it's relative to the root directory, not the current file).

`` :ref:`... <some-ref>` `` works like expected.

### DIFF
--- a/lib/Renderers/SpanNodeRenderer.php
+++ b/lib/Renderers/SpanNodeRenderer.php
@@ -168,7 +168,11 @@ abstract class SpanNodeRenderer implements NodeRenderer, SpanRenderer
                 $metaEntry = $this->environment->getMetaEntry();
 
                 if ($metaEntry !== null && $metaEntry->hasTitle($link)) {
-                    $url = $metaEntry->getUrl() . '#' . Environment::slugify($link);
+                    // A strangely-complex way to simply get the current relative URL
+                    // For example, if the current page is "reference/page", then
+                    // this would return "page" so the final URL is href="page#some-anchor".
+                    $currentRelativeUrl = $this->environment->relativeUrl('/' . $metaEntry->getUrl());
+                    $url                = $currentRelativeUrl . '#' . Environment::slugify($link);
                 }
             }
 
--- a/tests/Builder/BuilderTest.php
+++ b/tests/Builder/BuilderTest.php
@@ -400,7 +400,7 @@ class BuilderTest extends BaseBuilderTest
         $contents = $this->getFileContents($this->targetFile('subdir/test.html'));
 
         self::assertStringContainsString(
-            '<a href="subdir/test.html#em">em</a>',
+            '<a href="test.html#em">em</a>',
             $contents
         );
     }
--- a/tests/BuilderReferences/input/index.rst
+++ b/tests/BuilderReferences/input/index.rst
@@ -1,0 +1,11 @@
+References
+==========
+
+Test :ref:`referencing a document <reference>`
+
+Test :ref:`referencing an anchor in another document <test_anchor>`
+
+Test referencing a `Sub section`_ and `Some Sub Section`_ (invalid)
+
+Sub section
+-----------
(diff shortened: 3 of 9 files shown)


tests/Functional/tests/anchor/anchor.html
```diff
@@ -5,4 +5,6 @@ <h1>
     <a id="lists"></a>
 
     <p><a href="#lists">go to lists</a></p>
+    <p><a href="#lists">go to lists</a></p>
+    <p>Go to <a href="#anchors">Achors</a>.</p>
```
greg0ire:
```suggestion
    <p>Go to <a href="#anchors">Anchors</a>.</p>
```

tests/BuilderReferences/input/index.rst
```diff
@@ -0,0 +1,11 @@
+References
+==========
+
+Test :ref:`referencing a document <reference>`
```
wouterj, line 4: Fyi, this is officially not supported by the reSt spec. To reference documents, you must use ``:doc:``. But this was already supported, so I guess we have to keep it.

greg0ire:
Hi! This branch is non longer maintained: https://github.com/doctrine/rst-parser/blob/0aa99d5024010576400dab4efd84820da0acf799/.doctrine-project.json#L25-L30

wouterj:
Fixed :)

tests/Builder/BuilderTest.php
```diff
@@ -400,7 +400,7 @@ public function testReferenceToTitleWith2CharactersLong(): void
         $contents = $this->getFileContents($this->targetFile('subdir/test.html'));
 
         self::assertStringContainsString(
-            '<a href="subdir/test.html#em">em</a>',
+            '<a href="test.html#em">em</a>',
```
weaverryan, line 403: This, I believe, was actually WRONG previously. Here is the source file (it's `test.html` in the output due to the `url` directive on top): https://github.com/doctrine/rst-parser/blob/0.5.x/tests/Builder/input/subdir/index.rst

It makes sense that when this renders to `test.html` that, in order to link to a header in ITSELF, it would be just `href="test.html#em"`. This is precisely the bug this PR fixes... and I guess we had a test for the wrong behavior before.

weaverryan:
Bug fixed! This should be ready to go :)

lib/Renderers/SpanNodeRenderer.php
```diff
@@ -168,7 +168,11 @@ private function renderLink(SpanToken $spanToken, string $span): string
                 $metaEntry = $this->environment->getMetaEntry();
 
                 if ($metaEntry !== null && $metaEntry->hasTitle($link)) {
-                    $url = $metaEntry->getUrl() . '#' . Environment::slugify($link);
+                    // A strangely-complex way to simply get the current relative URL
+                    // For example, if the current page is "reference/page", then
+                    // this would return "page" so the final URL is href="page#some-anchor".
+                    $currentRelativeUrl = $this->environment->relativeUrl('/' . $metaEntry->getUrl());
+                    $url                = $currentRelativeUrl . '#' . Environment::slugify($link);
```
weaverryan, line 175: Unless I'm missing something, there is no way to simply ask:

> What is the relative URL to "myself" (the current file)?

We can ask `Environment` for the current filename (e.g. `reference/page`) ... and convert absolute paths (e.g. `/reference/page`) to relative URLs given the current filename... but there is nothing that does what I did above automatically.

greg0ire:
@wouterj please review the fix :)

wouterj:
My bad. I couldn't approve this PR and forgot to leave a comment. It's ready to go imho :+1: 

image

greg0ire (approved):

greg0ire:
Thanks @weaverryan @wouterj  !

PR #168, opened 2021-10-20
Fixing indentation stripped (affect diff highlighting)

weaverryan:
Hi!

This fixes https://github.com/symfony-tools/docs-builder/pull/124

tl;dr. Suppose you have

```
code-block: diff

      Normal Line
    + Added line
```

The code-block indentation is really 4 spaces. But because that first line (`Normal Line`) is indented by 6 spaces (2 of which should be part of the code block), the `Node` class currently thinks that the code block is indented by 6 spaces. It then strips the first 6 characters off each line, which strips the `+ ` in front of `+ Added line`.

Cheers!

### DIFF
--- a/lib/Nodes/Node.php
+++ b/lib/Nodes/Node.php
@@ -15,6 +15,7 @@ use Doctrine\RST\Renderers\NodeRendererFactory;
 use Doctrine\RST\Renderers\RenderedNode;
 
 use function implode;
+use function ltrim;
 use function strlen;
 use function substr;
 use function trim;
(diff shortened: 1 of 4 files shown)


tests/Functional/FunctionalTest.php
```diff
@@ -68,7 +70,7 @@ public function testFunctional(
 
         $rendered = $document->$renderMethod();
 
-        if ($format === Format::HTML) {
+        if ($format === Format::HTML && $useIndenter) {
```
weaverryan, line 73: This is the ugliest part of the PR: using the Indenter really made a mess of the HTML - it was removing some line breaks in odd places. Instead of making the HTML file match what the `Indenter` wanted, I removed it for this test. It makes it clear that we're getting the exact HTML we want.

lib/Nodes/Node.php
```diff
@@ -171,6 +176,17 @@ protected function getRenderer(): NodeRenderer
...
+    private function countStartingWhitespace(string $line): int
+    {
+        for ($k = 0; $k < strlen($line); $k++) {
+            if (trim($line[$k]) !== '') {
+                break;
+            }
+        }
+
+        return $k;
+    }
```
wouterj: Can you think of a reason why this for loop was used, instead of using `strlen($line) - strlen(ltrim($line))` to calculate the indent size?

tests/Functional/FunctionalTest.php
wouterj: Ref https://github.com/gajus/dindent/pull/28 (maybe we can somehow reactive that PR?)

wouterj (approved):
Looks good

lib/Nodes/Node.php
weaverryan: That seems to work fine 👍

greg0ire:
Thanks @weaverryan !

PR #173, opened 2021-11-09
Improve the way debug messages are output to STDERR

javiereguiluz:
I'm seeing this error occasionally on my local machine:

In case others are suffering from this too, I think we could improve code to not rely on this `STDERR` constant.

### DIFF
--- a/lib/Builder/ParseQueueProcessor.php
+++ b/lib/Builder/ParseQueueProcessor.php
@@ -66,8 +66,11 @@ final class ParseQueueProcessor
 
     private function processFile(string $file): void
     {
-        if (getenv('SHELL_VERBOSITY') >= 1 && $stdErr = fopen('php://stderr', 'wb')) {
-            fwrite($stdErr, sprintf("Processing file: %s\n", $file));
+        if (getenv('SHELL_VERBOSITY') >= 1) {
+            $stdErr = fopen('php://stderr', 'wb');
+            if (false !== $stdErr) {
+                fwrite($stdErr, sprintf("Processing file: %s\n", $file));
+            }
         }
 
         $fileAbsolutePath = $this->buildFileAbsolutePath($file);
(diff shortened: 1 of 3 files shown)


lib/Builder/ParseQueueProcessor.php
```diff
@@ -68,7 +67,10 @@ public function process(ParseQueue $parseQueue): void
     private function processFile(string $file): void
     {
         if (getenv('SHELL_VERBOSITY') >= 1) {
-            fwrite(STDERR, sprintf("Processing file: %s\n", $file));
+            $stdErr = fopen('php://stderr', 'wb');
+            if ($stdErr !== false) {
+                fwrite($stdErr, sprintf("Processing file: %s\n", $file));
```
greg0ire: You are getting this issue because you are in the HTTP context, right? In the case of HTTP, where is this output?

javiereguiluz: Yes, in the HTTP context. I don't know what happens with those messages. In that context I don't care much because I get the result as a value object with the list of error messages, etc.

In the CLI context I care about them, but I can get it easily.

Thanks!

greg0ire: Ok, so if the issue is that we are in the HTTP context, how about reverting this and checking `PHP_SAPI`?

lib/Builder/ParseQueueProcessor.php
```diff
@@ -68,7 +67,10 @@ public function process(ParseQueue $parseQueue): void
     private function processFile(string $file): void
     {
         if (getenv('SHELL_VERBOSITY') >= 1) {
```
greg0ire:
```suggestion
        if (getenv('SHELL_VERBOSITY') >= 1 && PHP_SAPI === 'cli') {
```

lib/Builder/ParseQueueProcessor.php
javiereguiluz: Yes, your solution looks better! I've updated the PR. Thanks.

greg0ire (approved):

greg0ire:
Thanks @javiereguiluz !

Issue #145, opened 2021-04-05
Add transformers/node visitors


wouterj:
It would be great if there is a supported way to traverse the generated `DocumentNode` and modify the node structure.

The [`docutils.transforms`](https://sourceforge.net/p/docutils/code/HEAD/tree/trunk/docutils/docutils/transforms/) namespace includes quite some usage of such node visitor, some directly applicable to this library:

* Using `PendingNodes` to better handle the `.. class::` directive (see discussion in https://github.com/doctrine/rst-parser/pull/141#discussion_r598263469 )
* Allow auto-numbering sections ([ref](https://sourceforge.net/p/docutils/code/HEAD/tree/trunk/docutils/docutils/transforms/parts.py#l18))
* Generating table of contents from a document node ([ref](https://sourceforge.net/p/docutils/code/HEAD/tree/trunk/docutils/docutils/transforms/parts.py#l70))

PR #175, opened 2022-01-09
Fix duplicate implicit references with the same id

wouterj:
This fixes a common bug in the docs (it has been reported multiple times over the past couple weeks).

If two or more titles on the same page have the same name, the references to the sections is wrong (as the ID is duplicate, you can never link to the second section).

Sphinx/docutils solve this by replacing any duplicate internal references with `id1`, `id2`, etc. I suggest making things a bit nicer by adding `-1`, `-2`, ... to the actual reference ID instead.

For reference, docutils actually manages references using a node visitor pattern. We've discussed adding something like this before (#145), but with the ongoing development on PHPdocumentor's side, I'm trying to keep the changes to this library to a minimal.

### DIFF
--- a/lib/Nodes/DocumentNode.php
+++ b/lib/Nodes/DocumentNode.php
@@ -13,6 +13,7 @@ use Exception;
 use function array_unshift;
 use function assert;
 use function count;
+use function in_array;
 use function is_string;
 use function sprintf;
 
@@ -32,6 +33,9 @@ class DocumentNode extends Node
 
     /** @var Node[] */
     private $nodes = [];
+
+    /** @var string[] */
+    private $implicitLinkTargets = [];
 
     public function __construct(Environment $environment)
     {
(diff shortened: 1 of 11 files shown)


wouterj:
Converting to draft: I realized some parts of this library still rely on simply calling `Environment::slugify()` to get a reference to a title.

lib/Nodes/DocumentNode.php
```diff
@@ -91,6 +95,24 @@ public function getNodes(?callable $function = null): array
         return $nodes;
     }
 
+    /**
+     * Creates an implicit hyperlink target for this document.
+     *
+     * @param string $preferredId The preferred ID of the hyperlink target, the actual ID is returned by the method (this avoids duplicates)
```
OskarStark, line 101:
```suggestion
     * @param string $preferredId The preferred ID of the hyperlink target, the actual ID is returned by the method to avoids duplicates
```
?

linawolf:
@wouterj might this be fixed by https://github.com/doctrine/rst-parser/pull/232?

PR #176, opened 2022-01-09
Rename "links" to "link targets"

wouterj:
This is something that took me a while to understand when working on #175. For me, "links" indicate the actual thing linking to something, not the thing that is linked to. The reStructured Text spec calls these "hyperlink targets", let's use that in this library as well.

TODO:
* [x] Update UPGRADE nodes
* [x] ~Do we want to implement a BC layer?~

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,18 @@
+# Upgrade to 0.6
+
+## Renamed links to link targets
+
+The library has been updated to use "link target" instead of "link":
+
+* Renamed `Environment::setLink()`, `Environment::getLink()` and
+  `Environment::getLinks()` to `Environment::setLinkTarget()`,
+  `Environment::getLinkTarget()` and `Environment::getLinkTargets()`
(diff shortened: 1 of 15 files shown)


greg0ire:
>  Do we want to implement a BC layer?

I don't think so: the release number starts with a `0.`

greg0ire (approved):

javiereguiluz (approved):

derrabus (approved):

wouterj:
Added the relevant information to the UPGRADE note. Ready to merge :)

greg0ire:
@wouterj doesn't this qualify as a bugfix?

wouterj:
The code doesn't change any behavior, it just changes method/properties names to better suit their actual function. Besides, it breaks BC for other libraries using these methods (just checked, symfony-tools/docs-builder doesn't use it)

derrabus:
It's a deliberate BC break and as such should go into a 0.x.0 release.

greg0ire:
Thanks! Merged!

PR #178, opened 2022-03-22
Allow Symfony 6

wouterj:
@greg0ire I think I need your help with 2 things here :)

* If I understand the GitHub workflow config correctly, we're now testing only Symfony 4 and Symfony 6. Is that OK for you, or do we need to also test Symfony 5? (if yes, please guide me through he process :smile: )
* I've targetted 0.6.x - the dev minor release. Does Doctrine consider adding support for newer dependency versions a bugfix (similar to Symfony)? I.e. should I target 0.5.x instead?

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -12,6 +12,6 @@ env:
 jobs:
   phpunit:
     name: "PHPUnit"
-    uses: "doctrine/.github/.github/workflows/continuous-integration.yml@1.1.0"
+    uses: "doctrine/.github/.github/workflows/continuous-integration.yml@1.4.1"
     with:
       php-versions: '["7.2", "7.3", "7.4", "8.0"]'
--- a/composer.json
+++ b/composer.json
@@ -19,10 +19,10 @@
     "require": {
         "php": "^7.2 || ^8.0",
         "doctrine/event-manager": "^1.0",
-        "symfony/filesystem": "^4.1 || ^5.0",
-        "symfony/finder": "^4.1 || ^5.0",
+        "symfony/filesystem": "^4.1 || ^5.0 || ^6.0",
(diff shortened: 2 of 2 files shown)


derrabus:
> If I understand the GitHub workflow config correctly, we're now testing only Symfony 4 and Symfony 6.

The job "PHPUnit (7.2, highest)" tests with Symfony 5 actually. That being said, shall we drop Symfony 4 soonish?

fabpot:
0.x is outside of semver anyway, so the version does not really matter.

greg0ire:
> shall we drop Symfony 4 soonish?

@derrabus maybe we should migrate https://github.com/doctrine/doctrine-website/blob/master/composer.json first?

greg0ire:
@wouterj my personal way is to reason in terms of stability instead of "is this a bugfix or not". For instance, new code is likely to make things unstable so, I would contribute it to the next minor.
But things like docs, CI improvements, or allowing new dependencies that are unlikely to cause bugs can and should target the next patch. Although that I agree that in this instance, it does not matter a lot, it would be cool if you could retarget to 0.5.x, for the extra reason that I will not have to create the 0.7.x branch.

It also would be great to release 1.0.0 at some point, so that we don't have to deal with the edge case of this library following 0.x versioning. Since both Doctrine and Symfony use this lib, it is IMO stable and we should avoid unnecessary BC-breaks. If that results in big major numbers, so be it IMO.

wouterj:
Alright, 0.5 it is :)

greg0ire (approved):

greg0ire:
The release workflow failed, let me fix it.

greg0ire:
https://github.com/doctrine/rst-parser/pull/179

greg0ire:
https://github.com/doctrine/rst-parser/releases/tag/0.5.2

wouterj:
Thank you @greg0ire !

PR #182, opened 2022-09-12
Add support for multi-line field options

wouterj:
Refs https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#field-lists


greg0ire:
Out of curiosity, why do you want to implement this? I suppose you want to improve Symfony's or Doctrine's docs, maybe a specific page?

lib/Parser/Directive.php
```diff
@@ -57,6 +57,6 @@ public function getOptions(): array
      */
     public function setOption(string $key, $value): void
     {
-        $this->options[$key] = $value;
+        $this->options[$key] = $value === '' ? true : $value;
```
greg0ire: How does this relate to your PR?

lib/Parser/DocumentParser.php
```diff
@@ -109,6 +109,9 @@ final class DocumentParser
     /** @var string|null */
     private $listMarker = null;
 
+    /** @var array{name: string,body: string,offset: int}|null */
```
greg0ire: We usually add a space after commas in such situations.

```suggestion
    /** @var array{name: string, body: string, offset: int}|null */
```

lib/Parser/LineDataParser.php
```diff
@@ -233,4 +233,20 @@ public function parseDefinitionList(array $lines): DefinitionList
 
         return new DefinitionList($definitionList);
     }
+
+    /**
+     * @return array{offset: int, name: string, body: string}
```
greg0ire:
```suggestion
     * @return array{offset: int, name: string, body: string}|null
```

tests/Parser/LineCheckerTest.php
```diff
@@ -91,4 +91,14 @@ public function testIsDefinitionListEnded(): void
...
         self::assertFalse($this->lineChecker->isDefinitionListEnded('', '    Definition'));
     }
+
+    public function testIsFieldList(): void
+    {
+        self::assertTrue($this->lineChecker->isFieldList(':glob:'));
+        self::assertTrue($this->lineChecker->isFieldList(':Date: 2001-08-16'));
+        self::assertTrue($this->lineChecker->isFieldList('    :Date: 2001-08-16'));
+        self::assertTrue($this->lineChecker->isFieldList(':Date published: 2001-08-16'));
+        self::assertTrue($this->lineChecker->isFieldList(':Date:published: 2001-08-16'));
```
greg0ire: > but colons (":") inside of field names must be backslash-escaped when followed by whitespace

That case does not seem to be tested. Also, the test determines whether there is a field list or not, but it does not say which part is the field name, and which is the body. Shouldn't there be a test for that?

wouterj:
> Out of curiosity, why do you want to implement this? I suppose you want to improve Symfony's or Doctrine's docs, maybe a specific page?

Oh, good question! :facepalm: 

I'm planning to add alt texts to all figures in the Symfony docs (ref https://github.com/symfony/symfony-docs/pull/16895 ). As these can be quite long, I want to use line breaks to conform to our soft 72 character limit in the Symfony docs.

wouterj:
Thanks for the review! I've refactored the changes a bit, to use a VO instead of the array structures. I also added the test cases you suggested.

lib/Parser/FieldOption.php
```diff
@@ -0,0 +1,64 @@
...
+
+    /** @var string */
+    private $body;
+
+    /** @var int */
+    private $lineCount = 0;
+
+    public function __construct(string $name, int $offset, string $body)
+    {
+        $this->name   = str_replace(' \:', ' :', $name);
```
greg0ire: >     but colons (":") inside of field names must be backslash-escaped when followed by whitespace

:thinking: it says "followed", is this really what should be done here? Or should `'\: ` be replaced with `: `?

tests/Parser/LineDataParserTest.php
```diff
@@ -52,6 +52,36 @@ public function getTestLinks(): array
...
+            self::assertNull($actual);
+        } else {
+            self::assertNotNull($actual);
+            self::assertSame($name, $actual->getName());
+            self::assertSame($body, $actual->getBody());
+        }
+    }
+
+    /**
+     * @return mixed[]
```
greg0ire:
```suggestion
     * @return array{string, string|null, true|string}
```

tests/Parser/LineDataParserTest.php
```diff
@@ -52,6 +52,36 @@ public function getTestLinks(): array
         ];
     }
 
+    /**
+     * @param mixed $body
```
greg0ire:
```suggestion
     * @param true|string $body
```

lib/Parser/FieldOption.php
wouterj: Haha, I finally understand the backslash thing now. Fixed it

greg0ire (approved):

greg0ire:
Thanks @wouterj !

PR #184, opened 2022-11-29
Specify `htmlspecialchars()` flags

greg0ire:
The test suite is broken on PHP 8.1 because the default value changed from `ENT_COMPAT` to `ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401`

See https://www.php.net/manual/en/function.htmlspecialchars.php

### DIFF
--- a/lib/HTML/Renderers/SpanNodeRenderer.php
+++ b/lib/HTML/Renderers/SpanNodeRenderer.php
@@ -13,6 +13,8 @@ use Doctrine\RST\Templates\TemplateRenderer;
 use function htmlspecialchars;
 use function trim;
 
+use const ENT_COMPAT;
+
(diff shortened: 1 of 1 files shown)


lib/HTML/Renderers/SpanNodeRenderer.php
```diff
@@ -69,7 +71,7 @@ public function link(?string $url, string $title, array $attributes = []): strin
 
     public function escape(string $span): string
     {
-        return htmlspecialchars($span);
+        return htmlspecialchars($span, ENT_COMPAT);
```
greg0ire, line 74: The alternative solution would be to use the new values and adapt the test, but it's kind of a BC-break, so maybe todo on 0.6.x? Let me know what you think.

derrabus (approved):

PR #193, opened 2022-12-01
[TASK] Create failed test for toctree with titlesonly option

### LINKED ISSUE #192: toctree ignores titlesonly option
When the option `:titlesonly:` is set the toctree may only contain the main titles of the documents, not the subtitles within. The main title of a document is defined to be the first one, no matter the syntax of underline / upperline

https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-toctree

I will provide a test

linawolf:


greg0ire:
Converting to draft until this is ready for review (it's not right now, right?)

greg0ire:
@linawolf sorry I should have said why I think the PR is not ready: the test errors out, presumably because the dataprovider now returns something that does not match the test function signature. I've been working on a PR to make this easier to spot: https://github.com/doctrine/rst-parser/pull/196

linawolf:
Locally the tests threw the expected results. So I just wait for #196 then work on this again?

greg0ire:
#196 won't make the tests fail instead of erroring out, but it should make PHPStan / Psalm point out the issue. So even if for some reason, phpunit works on your machine, PHPStan and Psalm won't.

tests/Functional/FunctionalTest.php
```diff
@@ -135,8 +165,16 @@ public function getFunctionalTests(): array
...
                     : 'render';
 
+                $renderMethod = in_array($basename, self::RENDER_ALL, true)
+                    ? 'renderAll'
+                    : $renderMethod;
+
                 $useIndenter = ! in_array($basename, self::SKIP_INDENTER_FILES, true);
 
+                if ($renderMethod === 'renderAll') {
+                    $rst      = file_get_contents($indexFilename);
```
greg0ire: This method can return `false` (and make the tests error out).

tests/Functional/FunctionalTest.php
```diff
@@ -94,9 +122,10 @@ public function getFunctionalTests(): array
         $tests = [];
 
         foreach ($finder as $dir) {
-            $rstFilename = $dir->getPathname() . '/' . $dir->getFilename() . '.rst';
-            if (! file_exists($rstFilename)) {
-                throw new Exception(sprintf('Could not find functional test file "%s"', $rstFilename));
+            $rstFilename    = $dir->getPathname() . '/' . $dir->getFilename() . '.rst';
+            $indexFilename  = $dir->getPathname() . '/index.rst';
+            if (! file_exists($rstFilename) && ! file_exists($indexFilename)) {
```
greg0ire: You are using an `or` in the error message, and should use the same here.

```suggestion
            if (! file_exists($rstFilename) || ! file_exists($indexFilename)) {
```

linawolf: just one of them two has to exist. Therefore and

greg0ire: Sorry you're right, it's or in in the message because it's "Could not find functional test file (x Or y)" which translates to "Could not find functional test file x and could not find functional test file y", so this makes sense 👍

greg0ire:
It seems that the "7.2 lowest" job is behaving as you expect, and the "7.2 highest" isn't… could it be you need to run `composer update`?

greg0ire:
Although it's preceded with this warning:

> PHP Warning:  file_get_contents(/home/runner/work/rst-parser/rst-parser/tests/Functional/tests/toctree-titlesonly/toctree-titlesonly.rst): failed to open stream: No such file or directory in /home/runner/work/rst-parser/rst-parser/tests/Functional/FunctionalTest.php on line 127

linawolf:
Working on that warning now

greg0ire:
Looks great :ok_hand:

tests/Functional/FunctionalTest.php
```diff
@@ -94,12 +118,16 @@ public function getFunctionalTests(): array
...
         foreach ($finder as $dir) {
-            $rstFilename = $dir->getPathname() . '/' . $dir->getFilename() . '.rst';
-            if (! file_exists($rstFilename)) {
-                throw new Exception(sprintf('Could not find functional test file "%s"', $rstFilename));
+            $rstFilename   = $dir->getPathname() . '/' . $dir->getFilename() . '.rst';
+            $indexFilename = $dir->getPathname() . '/index.rst';
+            if (file_exists($rstFilename)) {
+                $rst = file_get_contents($rstFilename);
+            } elseif (file_exists($indexFilename)) {
+                $rst = file_get_contents($indexFilename) ?? '';
```
greg0ire:
```suggestion
                $rst = file_get_contents($indexFilename) ?? '';
```

I'm attempting to rebase this on 0.5.x and the newly updated phpstan tells me that the left expression is not nullable… let me try dropping the `?? ''` part, as I'm under the impression you fixed the previous issue some other way.

linawolf: Makes sense as we check weather the file exists

greg0ire:
@linawolf I might have a fix but it seems too easy… is there a difference between specifying `titlesonly` and setting `maxdepth` to `1`?

linawolf:
Yes there is, titles only limits to the main title of each document, it would still show subdocuments however. While max levels only applies to the levels in total as far as I know

linawolf:
I will work on this on the train toonight

greg0ire:
That makes sense, although I'm not 100% sure what you mean by subdocument, that concept is almost not mentioned in the sphinx docs: https://www.sphinx-doc.org/en/master/search.html?q=subdocument
Can you maybe give me an example?

linawolf:
I will provide a test

greg0ire:
Thank you :pray:

linawolf:
MMh, it seems like your doctree is not working like sphinx would interpret it yet. Therefore I think we have to handle tests for maxdepth separately

greg0ire:
`maxdepth` is broken and we need to fix it first, is that it? :sweat:

linawolf:
Toctree without maxdepth doesn't work the way it should be either, sorry

linawolf:
Lets wait for #198 and maybe resolval of general toctree problems first

linawolf:
Now waiting for #199

linawolf:
To be continued in #200, not sure why I cant reopen this one

PR #194, opened 2022-12-01
Upgrade phpstan

greg0ire:


lib/Builder.php
```diff
@@ -132,15 +128,6 @@ public function getMetas(): Metas
...
 
-    public function getParseQueue(): ParseQueue
-    {
-        if ($this->parseQueue === null) {
-            throw new LogicException('The ParseQueue is not set until after the build is complete');
-        }
-
-        return $this->parseQueue;
-    }
-
```
SenseException, line 143: Removing a public method from a patch version? What's the connection to PHPStan?

greg0ire: `parseQueue` is never ever written to, the class is final and that property private, and that method is unused. It should have been removed in 655bd0133007f183de39c4e68c8bab16b6b8e0d8

SenseException (approved):

PR #195, opened 2022-12-01
Upgrade coding standard library

greg0ire:
I also configured the target PHP version instead of excluding rules one by one.

PR #198, opened 2022-12-04
[TASK] Create Functional Tests for Toctrees

linawolf:
I would suggest we start with a working test and get that one merged first, then we can take care of the other issues


linawolf:
I am not sure why the Coding Standards are failing here, locally `vendor/bin/phpcs` runs fine for me

greg0ire:
You may want to make sure you are rebased on the tip of  0.5.x, and run `composer update`, because of #195

tests/Functional/FunctionalTest.php
```diff
@@ -35,6 +36,7 @@ class FunctionalTest extends TestCase
 {
     private const RENDER_DOCUMENT_FILES = ['main-directive'];
     private const SKIP_INDENTER_FILES   = ['code-block-diff'];
+    private const RENDER_ALL            = ['toctree'];
```
greg0ire: I'm wondering why makes these functional tests special and deserving of more test code… my understanding is that since this is about a TOC, we need many documents.  Do we really need to render them all though? Since only `index.html` is added in these commit, whereas there are many rst files, we don't actually render them all.

tests/Functional/FunctionalTest.php
```diff
@@ -66,19 +68,41 @@ public function testFunctional(
             $this->expectExceptionMessage($expectedExceptionMessage);
         }
 
-        $document = $parser->parse($rst);
+        if ($renderMethod === 'renderAll') {
```
greg0ire: Both `renderDocument` and `render` are actual methods of the `DocumentNode` class. `renderAll` is not, so this is a bit confusing. Besides, an extra `if` means more complexity, and complexity in tests is something we should avoid (because we don't test the tests).

Maybe instead of a single `tests` directory, we should have  several of them, and then have several tests with several data providers? Something like `tests/render` and `tests/build` (since you are calling `Builder::build()`)?

greg0ire (commented):
I noticed some of your commits are verified, some other are not, and I think that might be because one of the email addresses you use in your commits is not associated with your github profile.
If that's on purpose you can ignore my comment, otherwise please associate your email address with your Github account, or change the email in your commits to an address already associated with it. If you do not
want to expose your personal email address, you may use `$userName@users.noreply.github.com`, that way we can still reach you through Github.
Additionally, you will get credit for this contribution on your Github profile.

tests/Functional/FunctionalTest.php
linawolf: yeah, why not. Just not sure how to implement that

tests/Functional/FunctionalTest.php
linawolf: > I noticed some of your commits are verified, some other are not, and I think that might be because one of the email addresses you use in your commits is not associated with your github profile. If that's on purpose you can ignore my comment, otherwise please associate your email address with your Github account, or change the email in your commits to an address already associated with it. If you do not want to expose your personal email address, you may use `$userName@users.noreply.github.com`, that way we can still reach you through Github. Additionally, you will get credit for this contribution on your Github profile.

Not sure how to do that, I always commit and push via PHP Storm. Tried to squash related commits though, maybe that somehow removed my email?

linawolf: How can you render a propper toctree without rendering all the documents? From What I understand `TocNodeRenderer` uses `ResolvedReference` which seem to equal a rendered page in some way

greg0ire: Ah you're right, I didn't realize we were rendering them all, but not comparing them all with an expected rendering.

tests/Functional/FunctionalTest.php
greg0ire: I will give it a try.

tests/Functional/FunctionalTest.php
greg0ire: > Not sure how to do that

Apparently your email address ends with `w-commerce.de`. Is that email address listed at https://github.com/settings/emails ?

tests/Functional/FunctionalTest.php
greg0ire: Here is how it could look: https://github.com/linawolf/rst-parser/pull/1

tests/Functional/FunctionalTest.php
linawolf: If we are thinking about toctrees, comparing them all is over-the top, however if we think about different directives to be tested like genindex.rst / `.. index:: ` directive we might have to test other files. So maybe make the file to be compared configurable somehow? We could of course compare the whole file tree but would then have many false negatives when something else changes

greg0ire: > So maybe make the file to be compared configurable somehow?

Maybe yeah. No sure how to get this one right though (have a list of files to compare or to avoid comparing, compare only when an html/tex counterpart is available, etc.).

linawolf: For the time being maybe the index.html is enough, if we need more features add them in the future?

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @linawolf !

Issue #197, opened 2022-12-04
toctree with 2 levels does not show second level


linawolf:
will provide a test for that one as well

linawolf:
https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-toctree

```
This directive inserts a “TOC tree” at the current location, using the individual TOCs (including “sub-TOC trees”) of the documents given in the directive body. Relative document names (not beginning with a slash) are relative to the document the directive occurs in, absolute names are relative to the source directory. A numeric maxdepth option may be given to indicate the depth of the tree; by default, all levels are included.
```

However in the rst-parser a toctree does not render more then one level of pages (does not render the “sub-TOC trees”)

Instead the maxdepth directive was applied to the additional titles found on pages of level 1

Amoifr:
I looked into this and I think it comes down to a semantic difference in what `maxdepth` and the tree levels mean.

Today, `TocNodeRenderer::buildLevel()` renders, for each document listed in the `toctree`, the tree of that page's own **headings** (h1, h2, ...) down to `maxdepth`. It does not pull in the **sub-toctrees** of those documents, so the tree is one document deep.

The Sphinx behavior you quoted is a **document** tree: the listed documents, then recursively the documents referenced by their own toctrees, with `maxdepth` counting document levels rather than in-page heading levels.

Moving to the Sphinx semantics would change the output of every `toctree` across all consumers (symfony-docs in particular), so before writing a PR I would love to pin the expected behavior down. Could you share a small concrete example, ideally as a fixture, something like:

- `index.rst` with `.. toctree:: :maxdepth: 2` listing `a`
- `a.rst` with its own `.. toctree::` listing `b`
- `b.rst`

and the exact HTML you would expect the root toctree to produce? That would let me build it against a clear target and check the impact on the existing fixtures.

Happy to implement it once the expected output is nailed down. Thanks!

PR #199, opened 2022-12-05
[TASK] Make a failed test for multilevel toctrees

linawolf:
Failed test to demonstrate #197


linawolf:
@greg0ire this is what I meant by subpages in a multi-level tree. Right now the behaviour of any toctree in the doctrine rst-parster is the same as if maxdepth=1 would be set. If the doctrine rst parser sets maxdepth=1 the behaviour is the same like if titlesonly was set in sphinx. However a change to this would be a breaking change. How do the projects feel about this? For us we really would need multi level menus as displayed in this test and as are common in sphinx

greg0ire:
> However a change to this would be a breaking change.

If that is so, then you should target 0.6.x

@SenseException @wouterj @weaverryan , do you maybe have some idea of how breaking implementing this would be to the Doctrine or Symfony docs?

weaverryan:
I'm not sure tbh. Symfony already has a lot of `:maxdepth: 1`, but I really am not sure. However, if it's a bug fix (regardless of if it's breaking), shouldn't we do it?

linawolf:
@weaverryan If we merge #200 you would only have to replace all `:maxdepth: 1` with `:titlesonly` which should be straight forward

wouterj:
If the change replicates what Sphinx is doing, this will not affect the Symfony docs. All toctree directives have been implemented when we were still using Sphinx IIRC.

wouterj:
A quick code search on the Doctrine organisation shows that most of them are also using either depth or maxdepth. Except from most of the Doctrine1 documentation and these:

* https://github.com/doctrine/DoctrineCacheBundle/blob/7b690d59c361e429c741628e7338eb5e150d479d/Resources/doc/index.rst
* https://github.com/doctrine/doctrine-laminas-hydrator/blob/b76cfa71d5e494a8a71ada4e8ea6454fc0fc56ba/docs/en/index.rst
* https://github.com/doctrine/DoctrineORMModule/blob/bb27271954a96461a611ace772973257c33bc83a/docs/en/index.rst
* https://github.com/doctrine/couchdb-odm/blob/9bf3d8ed9a649d91b6a66aa439b05a8a1f1ef7fd/docs/en/index.rst
* https://github.com/doctrine/DoctrineBundle/blob/d2088fc50494e4e7441fecca54732245a613eeb6/Resources/doc/index.rst
* https://github.com/doctrine/DoctrineMongoODMModule/blob/1592e4bd4c851634a7a72737186caaa1d446c151/docs/en/index.rst

greg0ire:
> shouldn't we do it ?

I'm just trying to assess if we should do it on 0.6.x or if 0.5.x is fine.

From what @wouterj is showing, I think it might be fine to stick with 0.5.x

wouterj:
I would do it in 0.6 personally. It isn't necessarily more work to release either new minor or patch version, right? So in that case, doing the safe semver approach seems better to me

greg0ire:
> It isn't necessarily more work to release either new minor or patch version, right?

Not that much more work, no. Let's follow your recommendation then. Converting to draft as a reminder.

greg0ire:
Here is the PR to refresh 0.6.x: https://github.com/doctrine/rst-parser/pull/203

SenseException:
> do you maybe have some idea of how breaking implementing this would be to the Doctrine or Symfony docs?

No idea. rst-parser didn't get many bigger changes that I could see as a reference.

linawolf:
Not working on it anymore

PR #204, opened 2022-12-07
[FEATURE] Introduce Contents directive

linawolf:


lib/Directives/Contents.php
```diff
@@ -0,0 +1,35 @@
...
+namespace Doctrine\RST\Directives;
+
+use Doctrine\RST\Nodes\Node;
+use Doctrine\RST\Parser;
+
+final class Contents extends Directive
+{
+    public function __construct()
+    {
+    }
```
greg0ire: :thinking: what is the point of that empty constructor?

lib/NodeFactory/NodeFactory.php
```diff
@@ -45,6 +46,9 @@ public function createDocumentNode(Environment $environment): DocumentNode;
      */
     public function createTocNode(Environment $environment, array $files, array $options): TocNode;
 
+    /** @param string[] $options */
+    public function createContentsNode(Environment $environment, DocumentNode $documentNode, array $options): ContentsNode;
```
greg0ire, line 50: This is technically a BC-break. A solution would be to document this with `@method` on the interface, then switch from `@method` to a real method signature on 0.6.x, or to target 0.6.x directly. I think the latter solution would work be best, it's more straightforward and we don't need that level of user-friendliness for that library.

lib/Nodes/ContentsNode.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\RST\Nodes;
+
+use Doctrine\RST\Environment;
+
+class ContentsNode extends Node
```
greg0ire, line 9: Should this class become `final`?

lib/Directives/Contents.php
linawolf: Leftover. Removed

lib/Nodes/ContentsNode.php
linawolf: The other nodes are not final so I figured they are not final by design?

lib/NodeFactory/NodeFactory.php
linawolf: To be honest I don't get this whole construct with the Interface? This makes adding any renderer a breaking change? but I don't mind to target 0.6.x

greg0ire: Yeah there's only one implementation, and having that big of an interface makes it near impossible to implement… the reasoning behind it is here: https://github.com/doctrine/rst-parser/pull/13#discussion_r221410489

@jwage , I see you have a use case for this, do you actually have a custom implementation of this? And if you do use wrapping, do you have a proxy method for each and every method in this interface?

lib/Nodes/ContentsNode.php
greg0ire: Let's suppose so.

linawolf:
@greg0ire So I changed the base branch to 0.6.x, how shall we proceed? can this be reviewed and then merged? Or is there like a certain date where development is started for version 0.6?

greg0ire:
There is no time constraint, the branch exists, so it's open for contributions :slightly_smiling_face:

lib/Utility/TitleLinkUtility.php
```diff
@@ -0,0 +1,105 @@
...
+    /** @param string[]|string $title */
+    private function generateAnchorFromTitle($title): string
+    {
+        $slug = is_array($title)
+            ? $title[1]
+            : $title;
+
+        return Environment::slugify($slug);
+    }
+}
```
greg0ire, line 105: Have you considered covering the lines codecov points at with tests?

linawolf: I didn't introduce the code, just moved it, but sure, Y not?

greg0ire: Oh sorry, didn't notice that. If you can that's great, but it's not a requirement.

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @linawolf !

PR #205, opened 2022-12-14
[TASK] Introduce Tests for TitleLinkUtility

linawolf:


greg0ire:
In case you are wondering why the coverage decreases: https://github.com/doctrine/rst-parser/pull/206

tests/Utility/TitleLinkUtilityTest.php
```diff
@@ -0,0 +1,283 @@
...
+    /** @var TitleLinkUtility */
+    private $titleLinkUtility;
+
+    protected function setUp(): void
+    {
+        $this->environment      = $this->createMock(Environment::class);
+        $this->titleLinkUtility = new TitleLinkUtility($this->environment, 999);
+    }
+
+    public function testBuildEmptyTitleLinksCreateEmptyTocItemArray(): void
```
greg0ire:
```suggestion
    public function testBuildEmptyTitleLinksCreatesEmptyTocItemArray(): void
```

tests/Utility/TitleLinkUtilityTest.php
```diff
@@ -0,0 +1,283 @@
...
+    {
+        $this->titleLinkUtility = new TitleLinkUtility($this->environment, 999);
```
greg0ire: This property has already been initialized in `setUp()`

linawolf: but there are tests for different header levels. Not sure how to handle that as i don't know in what order they are executed

greg0ire: I'm not sure I get it… is `TitleLinkUtility` stateful?

linawolf: yes, we pass the desired level in the constructor

greg0ire: Ah I think I get it now.

IMO you should use `$this->titleLinkUtility` when the level is 999, and a local variable such as `$titleLinkUtility` otherwise.

greg0ire (approved):

greg0ire:
Thanks @linawolf !

Issue #208, opened 2022-12-16
Intersphinx linking


linawolf:
TYPO3 docs heavily rely on intersphinx linking (https://docs.readthedocs.io/en/stable/guides/intersphinx.html / https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) We would also find it quite cool if doctrine and symfony would offer intersphinx links because then we could use them to link directly to sections without relying on file paths.

Should intersphinx linking be implemented as an additional package as is done in the phyton world?

greg0ire:
I think Symfony does have links to Doctrine for instance in https://github.com/symfony/symfony-docs/blob/d85ca3e24b83111c16bcae6e0f18690502969015/doctrine.rst
The opposite is also true, for instance in https://www.doctrine-project.org/projects/doctrine-phpcr-odm/en/latest/reference/basic-mapping.html#symfony-bundle

Cc @weaverryan @wouterj

linawolf:
The link from the symfony docs is not going to doctrine but to another page in symfony docs (from https://symfony.com/doc/current/doctrine.html) to (https://symfony.com/doc/current/doctrine/dbal.html) the link to the doctrine project is an external URL

The links from doctrine to symfony are also simple external links. These will neither adjust when the page is moved in the other documentation nor can I make automatic adjustments for versions etc, nor use the title.

In TYPO3 Third party extensions or specialized tutorials can use intersphinx linking. In our settings.cfg you can set shortcuts for all docs that you want to link to i.e:

```
t3coreapi      = https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/
t3tsref        = https://docs.typo3.org/m/typo3/reference-typoscript/11.5/en-us/
t3start        = https://docs.typo3.org/m/typo3/tutorial-getting-started/main/en-us/
```

I can then for example link to a tutorial:

```
Learn how to :ref:`Install TYPO3 with composer <t3start:composer>` (Tutorial Getting Started)!
```

If we release TYPO3 v12 LTS we can just switch the links to the newest version once and for all in the settings.cfg

greg0ire:
The link form the Symfony doc is going (among other things) to https://www.doctrine-project.org/projects/doctrine-orm/en/current/reference/basic-mapping.html , so it is going to Doctrine (and is an external URL, yes). So it could be changed into an intersphinx link, right?

To clarify, I wasn't saying Doctrine or Symfony are already using intersphinx links, I was saying both might have use cases for it.

[...]

PR #211, opened 2022-12-17
[Feature] Introduce event to intercept reference resolving

linawolf:
This event makes it possible to create an event that resolves intersphinx links of the format
```
:ref:`myproject:mylink`
```

for example using the event to override the resolved reference.

references https://github.com/doctrine/rst-parser/issues/208

### DIFF
--- a/docs/en/events.rst
+++ b/docs/en/events.rst
@@ -73,7 +73,7 @@ The events you can listen for are as follows:
 - ``PreNodeRenderEvent::PRE_NODE_RENDER`` - Dispatches a method named ``preNodeRender()`` before a node is rendered.
 - ``PostNodeRenderEvent::POST_NODE_RENDER`` - Dispatches a method named ``postNodeRender()`` after a node is rendered.
 - ``MissingReferenceResolverEvent::PRE_REFERENCED_RESOVED`` - Dispatches a method named
-  ``resolveMissingReference()`` if a reference cannot be resolved by the build-in methods. This event can
+  ``resolveMissingReference()`` if a reference cannot be resolved by the built-in methods. This event can
   be used to override the ``ResolvedReference`` returned by ``Resolver->resolve`` and thereby implement
   other means of references, like for example intersphinx links. The event can be listened to
   in implementing projects or extending packages. See
--- a/docs/en/events.rst
+++ b/docs/en/events.rst
@@ -72,7 +72,10 @@ The events you can listen for are as follows:
 - ``PostParseDocumentEvent::POST_PARSE_DOCUMENT`` - Dispatches a method named ``postParseDocument()`` after a node is parsed.
 - ``PreNodeRenderEvent::PRE_NODE_RENDER`` - Dispatches a method named ``preNodeRender()`` before a node is rendered.
 - ``PostNodeRenderEvent::POST_NODE_RENDER`` - Dispatches a method named ``postNodeRender()`` after a node is rendered.
-- ``PreReferenceResolvedEvent::PRE_REFERENCED_RESOVED`` - Dispatches a method named
-  ``preReferenceResolved()`` before a reference is resolved. If the ``$resolvedReference``
-  of the event is set to any non null value, resolving of references is stopped
-  and the ``$resolvedReference`` of the event is used instead.
+- ``MissingReferenceResolverEvent::PRE_REFERENCED_RESOVED`` - Dispatches a method named
+  ``resolveMissingReference()`` if a reference cannot be resolved by the build-in methods. This event can
+  be used to override the ``ResolvedReference`` returned by ``Resolver->resolve`` and thereby implement
+  other means of references, like for example intersphinx links. The event can be listened to
+  in implementing projects or extending packages. See
+  `Package Intersphinx, provided by the TYPO3 Documentation Team <https://github.com/TYPO3-Documentation/intersphinx>`__
+  for an example implementation.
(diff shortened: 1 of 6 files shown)


greg0ire:
> Introduce event to interfere reference resoleveable

Are you sure that's what you meant to write?

linawolf:
Not quite sure, sounds strange

greg0ire:
"intercept" or "interfere with" maybe? Now I think I get the meaning :slightly_smiling_face:

greg0ire:
> This event makes it possible to create an event

"to create an event listener"? If yes, where would/should that listener be implemented ?

linawolf:
We describe  on the same page in the section above on how to create event listeners...

greg0ire:
Sorry, it shows how little I know about this library… so I guess such listeners typically will be implemented in consuming projects, such as doctrine-website.

linawolf:
Well actually I am planning on writing an additional package that you can require within one of your consuming projects to enable intersphinx linking in the hope that doctrine and symfony would follow our lead

linawolf:
Here is a very basic implementation: https://github.com/TYPO3-Documentation/intersphinx It uses the new event to replace all intersphinx links to https://example.com links

greg0ire (approved):

linawolf:
So @wouterj could you have a look if we could merge this one so I can move forward with the implementation of the extension?

SenseException (changes requested):
Having tests would be nice. I'd like to see failing tests in the future if someone breaks something during development.

linawolf:
@SenseException that would be nice, not sure how to do that though. How could I test an event without actually having an event? The other events are also not tested...

lib/References/Resolver.php
```diff
@@ -15,6 +16,19 @@ public function resolve(
...
+
+        $preReferenceResolvedEvent = new PreReferenceResolvedEvent($environment, $data, $attributes);
+
+        $eventManager->dispatchEvent(
+            PreReferenceResolvedEvent::PRE_REFERENCED_RESOVED,
+            $preReferenceResolvedEvent
+        );
+
+        if ($preReferenceResolvedEvent->getResolvedReference() !== null) {
+            return $preReferenceResolvedEvent->getResolvedReference();
```
greg0ire: You could write an integration test for `Resolver`, with real `Environment`, `Configuration` and `EventManager` objects, and register a dummy event listener in it, that would change the url to something hardcoded, and check that the url is indeed changed.

The fact that other events are untested is a bit unfortunate, hopefully that's not because it's impossible to test :crossed_fingers:

wouterj:
Sphinx actually uses a "missing reference" hook for the intersphinx behavior: https://github.com/sphinx-doc/sphinx/blob/master/sphinx/ext/intersphinx.py#L637

Would it make sense to follow their logic, to be as compatible as possible? (order is important for overriding references locally) So first try to resolve the reference normally, then dispatch an event when the reference could not be resolved. If a listener manages to resolve the reference, use that one, otherwise return `null`.

linawolf:
@wouterj good idea

lib/References/Resolver.php
linawolf: done

linawolf:
@SenseException, provided Tests now. @wouterj I trigger the event now, when the reference could not be resolved, like it is done in sphinx. @greg0ire I tested this event now. I believe future events can be tested in a similar fashion.

@all Please review again.

docs/en/events.rst
```diff
@@ -72,3 +72,10 @@ The events you can listen for are as follows:
 - ``PostParseDocumentEvent::POST_PARSE_DOCUMENT`` - Dispatches a method named ``postParseDocument()`` after a node is parsed.
 - ``PreNodeRenderEvent::PRE_NODE_RENDER`` - Dispatches a method named ``preNodeRender()`` before a node is rendered.
 - ``PostNodeRenderEvent::POST_NODE_RENDER`` - Dispatches a method named ``postNodeRender()`` after a node is rendered.
+- ``MissingReferenceResolverEvent::PRE_REFERENCED_RESOVED`` - Dispatches a method named
+  ``resolveMissingReference()`` if a reference cannot be resolved by the build-in methods. This event can
```
greg0ire:
```suggestion
  ``resolveMissingReference()`` if a reference cannot be resolved by the built-in methods. This event can
```

tests/References/ResolverTest.php
```diff
@@ -120,4 +129,23 @@ public function testUnResolvedReference2(): void
 
         self::assertNull($this->resolver->resolve($this->environment, 'invalid-reference'));
     }
+
+    public function testMissingReferenceResolverEventDispatched(): void
+    {
+        $this->environment->expects(self::any())
```
greg0ire: This is a no-op, you can omit it. 

```suggestion
        $this->environment
```

tests/References/ResolverTest.php
```diff
@@ -120,4 +129,23 @@ public function testUnResolvedReference2(): void
...
+            ->method('getConfiguration')
+            ->willReturn($this->configuration);
+        $eventManager = new EventManager();
+        $eventManager->addEventListener(
+            [MissingReferenceResolverEvent::MISSING_REFERENCE_RESOLVER],
+            new SimpleMissingReferenceResolverListener()
+        );
+        $this->configuration->expects(self::any())
```
greg0ire:
```suggestion
        $this->configuration
```

greg0ire (commented):
AFK until early January so don't expect a very thorough review from me since I'm doing it from my phone. Thanks for adding tests!

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @linawolf!

Issue #210, opened 2022-12-17
Raise minimum PHP requirement to 7.4 for 0.6.x


linawolf:
This way we can use typed properties. Both the doctrine website and the symfony website use 7.4. We are planning to probably use 8.0 or 8.1 what do you think @greg0ire?

greg0ire:
I think [it looks sensible](https://packagist.org/packages/doctrine/rst-parser/php-stats)

PR #212, opened 2022-12-18
[Task] Drop PHP support below 7.4

linawolf:
This way we can use property types etc in the future

Due to this bug: https://github.com/twigphp/Twig/issues/3151 we have to use more recent versions of twig. The rest of the errors required a more recent version of phpunit

resolves https://github.com/doctrine/rst-parser/issues/210

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -29,4 +29,4 @@ jobs:
     name: "PHPUnit"
     uses: "doctrine/.github/.github/workflows/continuous-integration.yml@2.1.0"
     with:
-      php-versions: '["7.2", "7.3", "7.4", "8.0", "8.1", "8.2"]'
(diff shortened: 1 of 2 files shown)


linawolf:
@greg0ire Any Idea why this chnage fails the PHP unit test? Locally the tests are running with no issue

greg0ire:
I think that's because PHP 7.4 was never tested before with `--prefer-lowest`. One of our version constraints must be wrong. Try `composer update --prefer-lowest` locally, then you can try altering the version constraints.

linawolf:
@greg0ire I just did that, however phpunit still throws no errors...

linawolf:
Sorry, forget about it, I was running on PHP 8.0

linawolf:
@greg0ire could you have another look please?

greg0ire (approved):

derrabus (approved):

greg0ire:
Thanks @linawolf !

Issue #191, opened 2022-12-01
Introduce Text roles


linawolf:
Introduce Text roles that are not links

It seems like in the Doctrine rest parser any text role like

```
:something:`whatever
```
Is considered as link (therefore probably the name reference?)
However in the TYPO3 docs we use them frequently for special styles like
```
Click on :guilabel:`Save`
Use shortcurt :kbd:`Ctrl` :kbd:`S`
```

Compare https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#roles

PR #214, opened 2022-12-20
[FEATURE] Introduce Text Roles

linawolf:
resolves https://github.com/doctrine/rst-parser/issues/191

### DIFF
--- a/lib/Environment.php
+++ b/lib/Environment.php
@@ -10,6 +10,7 @@ use Doctrine\RST\NodeFactory\NodeFactory;
 use Doctrine\RST\References\Reference;
 use Doctrine\RST\References\ResolvedReference;
 use Doctrine\RST\Templates\TemplateRenderer;
+use Doctrine\RST\TextRoles\TextRole;
 use InvalidArgumentException;
 use Symfony\Component\String\Slugger\AsciiSlugger;
 
@@ -52,6 +53,9 @@ class Environment
 
     /** @var Reference[] */
     private $references = [];
+
+    /** @var TextRole[] */
+    private $textRoles = [];
 
     /** @var Metas */
     private $metas;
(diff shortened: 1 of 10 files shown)


lib/Environment.php
```diff
@@ -53,6 +54,9 @@ class Environment
     /** @var Reference[] */
     private $references = [];
 
+    /** @var TextRole[] */
```
greg0ire: Let's document the key type as well
```suggestion
    /** @var array<string, TextRole> */
```

lib/Parser.php
```diff
@@ -88,6 +89,15 @@ public function initDirectives(): void
         }
     }
 
+    public function initTextRoles(): void
+    {
+        $textRoles = $this->kernel->getTextRoles();
+
+        foreach ($textRoles as $textRole) {
```
greg0ire: Let's inline this.

```suggestion
        foreach ($this->kernel->getTextRoles() as $textRole) {
```

lib/Environment.php
linawolf: Someone ™️ should do the same for the references and directives...

greg0ire (approved):

SenseException (approved):
Approved, but is there a reason why `lib/TextRoles/TextRole.php` is an abstract class and not an interface?

linawolf:
> Approved, but is there a reason why `lib/TextRoles/TextRole.php` is an abstract class and not an interface?

My idea was to implement it the same way like abstract class Directive and abstract class Reference. So fututure changes would be less breaking.

wouterj (approved):
Looking good to me.

We currently use references for our custom text roles, but this seems much more clean. In the future, we might want to implement the reference feature of this component as a text role, but let's keep the changes and impact low for now :)

greg0ire:
Thanks @linawolf

Issue #187, opened 2022-11-30
Directives and Referencs are interpreted case-sensitive


linawolf:
While Sphinx interprets them case insensitive. Therefore in Spinx

```
.. important:: 
    Very important

and

.. Important:: 
    Very important
```` 

Bot work, in the doctrine parser only the first one works. Is this a bug or a feature?

PR #216, opened 2022-12-23
[BUGFIX] Ignore case in directive names

linawolf:
This might be a breaking change for some projects that relied on directives beeing case-sensitive. However Sphinx is not case sensitive so I would suggest to switch this.

references https://github.com/doctrine/rst-parser/issues/187


greg0ire (approved):

greg0ire:
Thanks @linawolf

PR #223, opened 2023-01-01
[BUGFIX] Ignore cache when running Builder functional tests

linawolf:


greg0ire:
Should this really be labelled as "bugfix"? It looks like it's more about making it easier to work on test. Either way, it's fine to merge this on 0.5.x

greg0ire (approved):

linawolf:
I don't think general tests (not about the cache) should have been run with cache on in the first place? But of course its a bug of the test, not of the project

PR #226, opened 2023-01-01
[BUGFIX] replace deprecated PHP unit method

linawolf:


linawolf:
This solution is not compatible with the lowest phpunit versions used in 0.5, lets only do this for 0.6

PR #228, opened 2023-01-01
[TASK] Compare lines in HTML functional tests trimmed

linawolf:
Adjusting the whitespace in HTML files to what our renderer outputs is tiresome.

And nice formating of the html is really not in the scope of this project.  Multiple whitespaces are ignored in HTML anyway. So I would suggest to trim the lines before comparing instead of rtrimming them.

Adjusted 2 Tests to demonstrate this is wokring.

A follow up which compares the HTML structure in a more logical way would be nice but having this already makes testing much easier without giving up acuracy

### DIFF
--- a/tests/Functional/FunctionalTest.php
+++ b/tests/Functional/FunctionalTest.php
@@ -24,7 +24,6 @@ use function file_get_contents;
 use function implode;
 use function in_array;
 use function is_string;
-use function rtrim;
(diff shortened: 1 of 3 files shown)


greg0ire (approved):

greg0ire:
Thanks @linawolf, and happy new year!

PR #231, opened 2023-01-06
[BUGIX] Do not register anonymous links

linawolf:
Anonymous links must not be registered as they have not name. Right now they are registered with their title.

This leads to multiple registrations of the same name eventually.

### DIFF
--- a/lib/Span/SpanProcessor.php
+++ b/lib/Span/SpanProcessor.php
@@ -188,7 +188,7 @@ final class SpanProcessor
         return $this->registerLinkFromMatches($match, true);
     }
 
-    /** @param String[] $match */
+    /** @param string[] $match */
     private function registerLinkFromMatches(array $match, bool $anonymous = false): string
     {
         $link = $match[3] !== '' ? $match[3] : $match[5];
--- a/lib/Span/SpanProcessor.php
+++ b/lib/Span/SpanProcessor.php
@@ -7,9 +7,7 @@ namespace Doctrine\RST\Span;
 use Doctrine\RST\Environment;
 use Doctrine\RST\Meta\LinkTarget;
 
-use function assert;
 use function htmlspecialchars;
-use function is_string;
 use function mt_rand;
 use function preg_match;
 use function preg_match_all;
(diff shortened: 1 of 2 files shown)


lib/Span/SpanProcessor.php
```diff
@@ -167,66 +165,76 @@ private function replaceLinks(string $span): string
...
-            // before setting the link and adding a token to be replaced
-            $link = str_replace("\n", ' ', $link);
-            $link = (string) preg_replace('/\s+/', ' ', $link);
+        return $span;
+    }
 
-            // we need to maintain the characters before and after the link
-            $prev = $match[1]; // previous character before the link
-            $next = $match[6]; // next character after the link
+    /** @param String[] $match */
```
greg0ire:
```suggestion
    /** @param string[] $match */
```

lib/Span/SpanProcessor.php
```diff
@@ -167,66 +165,76 @@ private function replaceLinks(string $span): string
...
+    private function registerAnonymousLinkFromMatches(array $match): string
+    {
+        return $this->registerLinkFromMatches($match, true);
+    }
 
-            $url = '';
+    /** @param String[] $match */
```
greg0ire:
```suggestion
    /** @param string[] $match */
```

greg0ire (commented):
I suppose this registry we have in the environment serves some purpose for named links, but not for anonymous links, and never did, right? Does this really fix an issue (beyond using memory for nothing), or is it just a refactoring?

linawolf:
Right now if you would first register a  target with the anchor `name` and then an anonymous link with the title `name` the latter would override the first withouth this expecting to happen.

I noted this while working on warnings for duplicate registered link targets. All the anonymous links were popping up as unexpected warnings.

greg0ire (approved):

tests/SpanProcessorTest.php
```diff
@@ -0,0 +1,35 @@
...
+
+    protected function setUp(): void
+    {
+        $this->environment = $this->createMock(Environment::class);
+    }
+
+    public function testProcessingLinkSpanSetsLinkTarget(): void
+    {
+        $this->environment->expects(self::exactly(1))->method('setLinkTarget');
+        $spanProcessor = new SpanProcessor($this->environment, '`example <http://example.org/>`_');
```
SenseException: I'd feel more secure about that test when `setLinkTarget` confirms that it gets the expected value.

SenseException (changes requested):

tests/SpanProcessorTest.php
linawolf: setLinkTarget gets called with an object of LinkTarget. I didn't find a way to test that and thought it is covered in the Functional tests then. Any idea?

linawolf: @SenseException , maybe like this with a callback?

SenseException: I think a simple instance creation of the object with the same expected constructor values should do the trick too. AFAIK objects are compared with `==` in `with()`
https://3v4l.org/GBLYU

linawolf: Cool, didn't know that

SenseException (approved):

SenseException:
Thank you @linawolf

PR #232, opened 2023-01-06
[FEATURE] Register link anchors globally in Metas

linawolf:
And warn about multiply defined link targets.

Right now anonymous links with the same title trigger warnings about duplicate link targets, this is resolved by https://github.com/doctrine/rst-parser/pull/231

### DIFF
--- a/lib/Meta/Metas.php
+++ b/lib/Meta/Metas.php
@@ -5,8 +5,15 @@ declare(strict_types=1);
 namespace Doctrine\RST\Meta;
 
 use Doctrine\RST\Environment;
+use Doctrine\RST\ErrorManager;
+use Doctrine\RST\UrlGenerator;
 
+use function array_key_exists;
+use function array_merge;
+use function serialize;
 use function strtolower;
+use function trim;
+use function unserialize;
 
 class Metas
 {
@@ -16,10 +23,20 @@ class Metas
     /** @var string[] */
     private $parents = [];
 
-    /** @param MetaEntry[] $entries */
-    public function __construct(array $entries = [])
+    /** @var array<string, LinkTarget> */
+    private array $linkTargets = [];
+
+    private ErrorManager $errorManager;
+
+    /**
+     * @param MetaEntry[]               $entries
+     * @param array<string, LinkTarget> $linkTargets
+     */
+    public function __construct(ErrorManager $errorManager, array $entries = [], array $linkTargets = [])
     {
-        $this->entries = $entries;
+        $this->errorManager = $errorManager;
+        $this->entries      = $entries;
+        $this->linkTargets  = $linkTargets;
     }
 
     public function findLinkTargetMetaEntry(string $linkTarget): ?MetaEntry
@@ -83,6 +100,8 @@ class Metas
         }
 
         $this->entries[$file]->setParent($this->parents[$file]);
+
+        $this->linkTargets = array_merge($this->linkTargets, $linkTargets);
     }
 
     public function get(string $url): ?MetaEntry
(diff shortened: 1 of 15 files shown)


linawolf:
@greg0ire do you have any idea, why the unit tests work locally but not on here?

I tried running them with PHP 7.4 and lowest and they still do not fail for me in ddev...

greg0ire:
I have no idea but I reproduce the issue locally with php 8.1

greg0ire:
Let me know if I can debug this somehow.

lib/Meta/Metas.php
```diff
@@ -125,4 +144,72 @@ private function findByTitle(string $text): ?MetaEntry
...
+            $i = 2;
+            while (array_key_exists($name . '-' . $i, $this->linkTargets)) {
+                $i++;
+            }
+
+            $name .= '-' . $i;
+            $linkTarget->setName($name);
+            $linkTarget->setDuplicate(true);
+        }
+
```
greg0ire, line 167: This part of the code is reached during the failing test, hope this helps.

linawolf:
I am getting insane here. I cloned the repository new, did a composer update, run the unit tests and they are still green. Switch the PHP version: still green. Do composer update with lowest or highest version: still green. Why can't I reproduce this locally?

lib/Meta/Metas.php
linawolf: ty

linawolf:
So it seems like we cannot rely on the order these documents are processed. As the rule is only: when there are two same anchors one is renamed and there is a warning, I just checked in both documents for the anchor, one has to be renamed. Still I think it is strange that the execution order is different on different machines, but when you repeately run the tests it always stays the same?

greg0ire:
All I know is you are not insane, for I cannot reproduce the issue on my home machine, only my work machine. I know PHP has some similar weirdnesses, for instance `get_declared_classes()` doesn't guarantee a particular order. So my guess is we miss a call to `sort()` somewhere.

greg0ire:
Also, this test seems to pollute the output through the error manager, do you also see warnings in the output?

lib/Meta/Metas.php
```diff
@@ -125,4 +144,72 @@ private function findByTitle(string $text): ?MetaEntry
...
+        if (array_key_exists($name, $this->linkTargets)) {
+            $this->errorManager->warning('Duplicate anchor ".. _' . $linkTarget->getName() . '" found.');
+            $i = 2;
+            while (array_key_exists($name . '-' . $i, $this->linkTargets)) {
+                $i++;
+            }
+
+            $name .= '-' . $i;
+            $linkTarget->setName($name);
+            $linkTarget->setDuplicate(true);
```
greg0ire, line 165: If I dump `$this->linkTargets` right before this `if` closes, I have different results:

In the case where it passes, the title is "Headline one", otherwise it's "Headline two", so it looks like you're right about this being an issue with the processing order.

linawolf:
> Also, this test seems to pollute the output through the error manager, do you also see warnings in the output?

The warnings are solved by https://github.com/doctrine/rst-parser/pull/231

lib/Meta/Metas.php
linawolf: All the documents are put into a queue for processing. Maybe on some machines these run on parallel processes or something like that? Anyhow we do not promise a certain order and renaming a link target that caused a warning here.

greg0ire: And indeed, when I dump `$file` inside `ParseQueueProcessor::process()`, I get either `index` or `duplicateAnchor`.

linawolf: So should this issue concern us? Is it a bug or a feature?

greg0ire: But as it turns out, this might not be an order issue because when I dump `$this->fileInfos` after the foreach loop in `Scanner::scan()`, I get 1 file vs 2: `duplicateAnchor` is not found when the test passes.

linawolf: in the output folder however duplicateAnchor.html gets generated and I explicitely load it in the test

linawolf: I do have both files in the Scanner:scan() $this->fileInfos after the for loop and I have passing tests...

greg0ire: Ah my work computer was not up to date with the latest changes. Now it works…

greg0ire: Oh but it works on the CI too, I didn't even notice that :facepalm:

greg0ire:
> Still I think it is strange that the execution order is different on different machines, but when you repeately run the tests it always stays the same?

Maybe it's a matter of running `git clean -fdx`? A lot of stuff persists after a test, it seems.

greg0ire:
> I cloned the repository new

But I guess you already tried that.

greg0ire:
> Still I think it is strange that the execution order is different on different machines, but when you repeately run the tests it always stays the same?

I find it strange as well, let me know if you want me to `var_dump()` some values anywhere in the code.

linawolf:
Well in this ticket I didn't change anything in the scanner or parser queue, as we didn't have warnings for duplicate anchors before, none would notice the execution order. Then question is: Do we care about execution order? I don't think so. We first do the parsing and resolve the links and such in a second step, the rendering. 

Now By renaming the second occurence of duplicated anchor we do create a difference in execution order here. However as there is a warning displayed anyway you cannot really expect for which one of the duplicate anchors the link was intended, so it might be - as well - one of them depending on execution order (and caching as well if that is enabled)

greg0ire:
Sounds fair… let's leave it as is then.

greg0ire (approved):

SenseException (approved):
I haven't run into any issues with the tests too. Let's see how the piece works with the whole puzzle.

greg0ire:
Thanks @linawolf !

PR #234, opened 2023-01-07
[Feature] Replace interpreted text

linawolf:
Interpreted text without explicit text role should be interpreted as default role.

In software projects the default role is usually code.

Right now these are interpreted as references that cannot be resolved causing warnings.

Compare https://docutils.sourceforge.io/docs/ref/rst/roles.html

### DIFF
--- a/tests/Functional/tests/render/interpreted-text/interpreted-text.rst
+++ b/tests/Functional/tests/render/interpreted-text/interpreted-text.rst
@@ -1,0 +1,6 @@
+Testing some `interpreted text`...
+
+`a` or `aa` or `.\/$!`
+
+`This long interpreted text is longer
+then one line`
--- a/tests/Functional/tests/render/interpreted-text/interpreted-text.rst
+++ b/tests/Functional/tests/render/interpreted-text/interpreted-text.rst
@@ -3,4 +3,4 @@ Testing some `interpreted text`...
 `a` or `aa` or `.\/$!`
 
 `This long interpreted text is longer
-then one line`
+than one line`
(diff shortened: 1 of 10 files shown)


tests/Functional/tests/render/interpreted-text/interpreted-text.rst
```diff
@@ -0,0 +1,6 @@
+Testing some `interpreted text`...
+
+`a` or `aa` or `.\/$!`
+
+`This long interpreted text is longer
+then one line`
```
greg0ire:
```suggestion
than one line`
```

greg0ire (commented):
The document you linked to states that the "default default role" is `:title-reference:`… is it legit to change that? Or should application be resorting to `default-role` to do so. Code does seem like a sensible default, but can the "default default role" really be changed?

linawolf:
Any project who would rather have the title as default default role could just override the templates...

greg0ire (approved):

tests/Functional/tests/render/interpreted-text/interpreted-text.rst
```diff
@@ -0,0 +1,6 @@
+Testing some `interpreted text`...
+
+`a` or `aa` or `.\/$!`
```
SenseException, line 3: Isn't code usually done with a double ```` in rst?  http://tpcg.io/_NZX3UJ

linawolf: Text in single backticks is considered "interpreted text" with the default "textrole". In Sphinx as well as in Githubs interpretation of RST, the default text-role is `<code>`. The only advantage of double backticks being, that you can use single backticks within, without escaping them. 

In RST-standard the default default-textrole is "title", a default which is less helpful in Documentation and is usually used in print. But each implementation may define its own fallback.

SenseException (approved):

PR #235, opened 2023-01-08
[FEATURE] Add support for standard rst and sphinx text roles

### LINKED ISSUE #78: missing support for many inline markup items - like :kbd:
We know about `:ref:`and `:doc:`, but there are actually a bunch of other `:something:` inline syntaxes supported by Sphinx - http://www.sphinx-doc.org/es/stable/markup/inline.html#cross-referencing-documents

We use at least one in the symfony docs - `:kbd:`. I don't think that we should go and implement *every* one of these (there are a lot) - but it looks like we *do* need some systematic way of handling this `:something:` syntax. If I'm not mistaken, it looks like we currently treat *anything* matching `:*:` as a reference: https://github.com/doctrine/rst-parser/blob/8e8aa20eb072e69f557e4a49c4db0b50928a4ef6/lib/Span/SpanProcessor.php#L107

linawolf:
Make standard wrapper text roles that only need to be instantiated, add example HTML markup. Allow textrole registration by Format

resolves #78

See https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/WritingReST/Reference/Code/InlineCode.html

https://docutils.sourceforge.io/docs/ref/rst/roles.html#toc-entry-3

### DIFF
--- a/lib/HTML/HTMLFormat.php
+++ b/lib/HTML/HTMLFormat.php
@@ -12,6 +12,9 @@ use Doctrine\RST\Renderers;
 use Doctrine\RST\Renderers\CallableNodeRendererFactory;
 use Doctrine\RST\Renderers\NodeRendererFactory;
 use Doctrine\RST\Templates\TemplateRenderer;
+use Doctrine\RST\TextRoles\DefinitionTextRole;
+use Doctrine\RST\TextRoles\TextRole;
+use Doctrine\RST\TextRoles\WrapperTextRole;
 
 final class HTMLFormat implements Format
 {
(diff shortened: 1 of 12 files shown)


greg0ire (approved):

lib/HTML/HTMLFormat.php
```diff
@@ -44,6 +47,28 @@ public function getDirectives(): array
...
+            new WrapperTextRole('kbd', '<kbd class="kbd docutils literal notranslate">%s</kbd>'),
+            new WrapperTextRole('mailheader', '<em class="mailheader">%s</em>'),
+            new WrapperTextRole('math', '<math>%s</math>'),
+            new WrapperTextRole('emphasis', '<em>%s</em>'),
+            new WrapperTextRole('literal', '<span class="pre">%s</span>'),
+            new WrapperTextRole('strong', '<em class="strong">%s</em>'),
+            new WrapperTextRole('subscript', '<sub>%s</sub>', ['sub']),
+            new WrapperTextRole('superscript', '<sup>%s</sup>', ['sup']),
+            new WrapperTextRole('title-reference', '<cite>%s</cite>', ['t', 'title']),
+            new DefinitionTextRole('abbreviation', '<abbr title="%2$s">%1$s</abbr>', ['abbr']),
```
SenseException: Where does the HTML representation of the roles come from? Did I miss something in the links?

SenseException (commented):
This PR looks fine and ready to merge, but there's one point about the HTML I don't understand. Wrote it in a review-comment.

lib/HTML/HTMLFormat.php
linawolf: RST does not specify any HTML representations as RST can be transformed into many different Formats. So I took those we are doing in TYPO3 or you have a better idea?

SenseException: Most of the tags make sense as they are the equivalent of the role. I'm confused about `<em class="strong">%s</em>` for strong. I'd expect the `<strong>` tag.

lib/HTML/HTMLFormat.php
```diff
@@ -44,6 +47,28 @@ public function getDirectives(): array
...
+            new WrapperTextRole('command', '<strong class="command">%s</strong>'),
+            new WrapperTextRole('dfn', '<em class="dfn">%s</em>'),
+            new WrapperTextRole('file', '<span class="pre file">%s</span>'),
+            new WrapperTextRole('guilabel', '<span class="guilabel">%s</span>'),
+            new WrapperTextRole('kbd', '<kbd class="kbd docutils literal notranslate">%s</kbd>'),
+            new WrapperTextRole('mailheader', '<em class="mailheader">%s</em>'),
+            new WrapperTextRole('math', '<math>%s</math>'),
+            new WrapperTextRole('emphasis', '<em>%s</em>'),
+            new WrapperTextRole('literal', '<span class="pre">%s</span>'),
+            new WrapperTextRole('strong', '<em class="strong">%s</em>'),
```
linawolf:
```suggestion
            new WrapperTextRole('strong', '<strong>%s</strong>'),
```

lib/HTML/HTMLFormat.php
linawolf: thats true, changed it to stron

greg0ire (approved):

SenseException (approved):

SenseException:
@linawolf Thank you

PR #236, opened 2023-01-08
[BUGFIX] Avoid NullPointer exception

linawolf:
Calling an unkown directive with field options causes a Nullpointer exception. We cannot rely on it to be set here. Unknown directives already cause warning


lib/Parser/DocumentParser.php
```diff
@@ -505,7 +505,7 @@ private function parseLine(string $line): bool
                 }
 
                 if ($this->lineChecker->isFieldOption($line)) {
-                    if ($this->fieldOption !== null) {
+                    if ($this->directive !== null && $this->fieldOption !== null) {
```
greg0ire, line 508: So we silently don't set the option? Should maybe an error end up in the error manager or something?

linawolf: Non-existent directives already throw warnings. I don't see a reason, why a directive that has field options should throw a stronger errror then a directive without. Non-existent directive is non-existent directive in my eyes?

greg0ire: I wasn't aware warnings were already thrown. In that case, this should be fine 👍

greg0ire (approved):

SenseException:
Please cover this case with a test. :test_tube:

linawolf:
@SenseException here you go

SenseException (changes requested):
The test files don't make phpunit fail when I revert to the old code. The tests are supposed to fail on `if ($this->fieldOption !== null) {` and work on `if ($this->directive !== null && $this->fieldOption !== null) {`

linawolf:
@SenseException, Ok, should be fixed now. The whole thing is a little tricky as weather or not the Nullpointer exception is reached depends on the settings of the error manager: If the error manager throws an Exception on finding the error of the non-existent directive then we never get to the code where the document parser tries to set the fields on the non-existent directive.

So this error only happens when you disable throwing exceptions on finding errors.

SenseException (approved):
Well done :+1:

SenseException:
Once again, thank you @linawolf

PR #237, opened 2023-01-09
[FEATURE] Introduce Pre and Post rendering document events

linawolf:
I am using the pre event to pass additional information, for example the menu structure to my template. The post event could be used for post processing for example.

### DIFF
--- a/docs/en/events.rst
+++ b/docs/en/events.rst
@@ -79,7 +79,7 @@ The events you can listen for are as follows:
   in implementing projects or extending packages. See
   `Package Intersphinx, provided by the TYPO3 Documentation Team <https://github.com/TYPO3-Documentation/intersphinx>`__
   for an example implementation.
-- ``PreDocumentRenderEvent``: This event is called in the DocumentNodeRenderer
+- ``PreDocumentRenderEvent``: This event is called in the `DocumentNodeRenderer`
   before the document Node is rendered. The event can be used to influence the
   parameters sent to the template and or the name of the template to be used
   during rendering.
--- a/docs/en/events.rst
+++ b/docs/en/events.rst
@@ -83,7 +83,7 @@ The events you can listen for are as follows:
   before the document Node is rendered. The event can be used to influence the
   parameters sent to the template and or the name of the template to be used
   during rendering.
-- ``PostDocumentRenderEvent``: This event is called in the DocumentNodeRenderer
+- ``PostDocumentRenderEvent``: This event is called in the `DocumentNodeRenderer`
   after the content of the document got rendered and before the HTML is
   returned. It can be used to postprocess or exchange the HTML that will be
   written to the file.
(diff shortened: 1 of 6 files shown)


docs/en/events.rst
```diff
@@ -79,3 +79,11 @@ The events you can listen for are as follows:
   in implementing projects or extending packages. See
   `Package Intersphinx, provided by the TYPO3 Documentation Team <https://github.com/TYPO3-Documentation/intersphinx>`__
   for an example implementation.
+- ``PreDocumentRenderEvent``: This event is called in the DocumentNodeRenderer
```
greg0ire: I'd either use "in `DocumentNodeRendererer`" (with code markup), or "in the document node renderer" here and in similar cases below.

lib/Event/PostDocumentRenderEvent.php
```diff
@@ -0,0 +1,83 @@
...
+use Doctrine\RST\Configuration;
+use Doctrine\RST\Nodes\DocumentNode;
+use Doctrine\RST\Templates\TemplateRenderer;
+
+use function sprintf;
+
+/**
+ * This event is called in the DocumentNodeRenderer after the content of
+ * the document got rendered and before the HTML is returned. It can be used
+ * to postprocess or exchange the HTML that will be written to the file.
```
greg0ire: Did you really mean to write "exchange"? To me, you exchange something with something else.

```suggestion
 * to postprocess or change the HTML that will be written to the file.
```

greg0ire (approved):

docs/en/events.rst
linawolf:
```suggestion
- ``PreDocumentRenderEvent``: This event is called in the `DocumentNodeRenderer`
```

docs/en/events.rst
```diff
@@ -79,3 +79,11 @@ The events you can listen for are as follows:
   in implementing projects or extending packages. See
   `Package Intersphinx, provided by the TYPO3 Documentation Team <https://github.com/TYPO3-Documentation/intersphinx>`__
   for an example implementation.
+- ``PreDocumentRenderEvent``: This event is called in the `DocumentNodeRenderer`
+  before the document Node is rendered. The event can be used to influence the
+  parameters sent to the template and or the name of the template to be used
+  during rendering.
+- ``PostDocumentRenderEvent``: This event is called in the DocumentNodeRenderer
```
linawolf:
```suggestion
- ``PostDocumentRenderEvent``: This event is called in the `DocumentNodeRenderer`
```

docs/en/events.rst
```diff
@@ -79,3 +79,11 @@ The events you can listen for are as follows:
   in implementing projects or extending packages. See
   `Package Intersphinx, provided by the TYPO3 Documentation Team <https://github.com/TYPO3-Documentation/intersphinx>`__
   for an example implementation.
+- ``PreDocumentRenderEvent``: This event is called in the `DocumentNodeRenderer`
```
greg0ire, line 82: Shouldn't you be using double backticks? Since this isn't markdown?

linawolf: As our default textrole is "code" the result is exactly the same: https://github.com/doctrine/rst-parser/pull/234

Also GiThub renders it like code as well. The only advantage of double backticks then is that you can use single backticks within without escaping them

SenseException (approved):

PR #238, opened 2023-01-09
[TASK] Refactor Error Management

linawolf:
* ErrorHandler is an interface now
* ErrorHandler is created by Factory
* The DefaultErrorHandler should exist only once per Configuration
* Configuration is responsible for delivering the ErrorHandler
* You can get your ErrorHandler only from the configuration, no other class should expose it so we have a single point of truth
* You can register your own ErrorHandler by using another ErrorHandlerFactory
* You can mock the ErrorHandlerFactory to test for expected / non-expected warnigns and errors
* Adjust Builder Tests to never expect warnings and errors unless configured to expect errors.

### DIFF
--- a/lib/ErrorManager/DefaultErrorManager.php
+++ b/lib/ErrorManager/DefaultErrorManager.php
@@ -1,0 +1,70 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\RST\ErrorManager;
+
+use Doctrine\RST\Configuration;
+use Doctrine\RST\Error;
+use Doctrine\RST\ErrorManager;
+use Exception;
+use Throwable;
+
+use function sprintf;
+
+final class DefaultErrorManager implements ErrorManager
+{
+    /** @var Configuration */
+    private $configuration;
+
+    /** @var list<Error> */
+    private $errors = [];
+
+    public function __construct(Configuration $configuration)
+    {
+        $this->configuration = $configuration;
+    }
+
+    private function log(string $level, string $message, ?string $file = null, ?int $line = null, ?Throwable $throwable = null): void
+    {
+        $this->errors[] = $error = new Error($level, $message, $file, $line, $throwable);
+        if (! $this->configuration->isSilentOnError()) {
+            if ($level === Error::LEVEL_WARNING && $this->configuration->isWarningsAsError()) {
+                $level = Error::LEVEL_ERROR;
+            }
+
+            if ($this->configuration->getOutputFormat() === Configuration::OUTPUT_FORMAT_GITHUB) {
+                $file = $error->getFile();
+                echo sprintf(
+                    '::%s %s%s::%s',
+                    $level,
+                    $file !== null ? 'file=' . $file : '',
+                    $file !== null && $error->getLine() !== null ? ',linefile=' . $error->getLine() : '',
+                    $error->getMessage()
+                );
+            } else {
+                echo ($level === Error::LEVEL_ERROR ? '⚡️ ' : '⚠️ ') . $error->asString() . "\n";
+            }
+        }
+
+        if ($level === Error::LEVEL_ERROR && $this->configuration->isAbortOnError()) {
+            throw new Exception($error->asString(), 0, $error->getThrowable());
+        }
+    }
+
+    public function error(string $message, ?string $file = null, ?int $line = null, ?Throwable $throwable = null): void
+    {
+        $this->log(Error::LEVEL_ERROR, $message, $file, $line, $throwable);
+    }
+
+    public function warning(string $message, ?string $file = null, ?int $line = null, ?Throwable $throwable = null): void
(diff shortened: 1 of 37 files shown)


lib/ErrorManager/DefaultErrorManager.php
```diff
@@ -0,0 +1,82 @@
...
+            if ($this->configuration->getOutputFormat() === Configuration::OUTPUT_FORMAT_GITHUB) {
+                $file = $error->getFile();
+                echo sprintf(
+                    '::error %s%s::%s',
+                    $file !== null ? 'file=' . $file : '',
+                    $file !== null && $error->getLine() !== null ? ',linefile=' . $error->getLine() : '',
+                    $error->getMessage()
+                );
+            } else {
+                echo '⚠️ ' . $error->asString() . "\n";
```
greg0ire: Maybe we should challenge having a warning sign for errors, and no sign for warnings.

tests/BaseBuilderTest.php
```diff
@@ -5,29 +5,61 @@
...
 use function shell_exec;
 
 abstract class BaseBuilderTest extends TestCase
 {
-    /** @var Builder */
-    protected $builder;
+    protected Configuration $configuration;
+    protected Builder $builder;
+    /** @var ErrorManager|MockObject */
+    protected $errorManger;
```
greg0ire: :yum: https://dict.leo.org/franz%C3%B6sisch-deutsch/manger

```suggestion
    protected $errorManager;
```

lib/ErrorManager/DefaultErrorManager.php
linawolf: This follows the general conception with programmers that warnings can be ignored and errors must be noted but can be ignored unless they blow everything up?

lib/ErrorManager/DefaultErrorManager.php
```diff
@@ -0,0 +1,82 @@
...
+
+        $error = new Error($message, $file, $line, $throwable);
+        if ($this->configuration->getOutputFormat() === Configuration::OUTPUT_FORMAT_GITHUB) {
+            $file = $error->getFile();
+            echo sprintf(
+                '::warning %s%s::%s',
+                $file !== null ? 'file=' . $file : '',
+                $file !== null && $error->getLine() !== null ? ',linefile=' . $error->getLine() : '',
+                $error->getMessage()
+            );
```
greg0ire: Should this be extracted in a private method? The only difference between this block and the one above seems to be the level.

tests/BaseBuilderTest.php
```diff
@@ -5,29 +5,61 @@
...
+        $this->configureExpectedWarnings();
+        $this->configureExpectedErrors();
+        $this->builder = new Builder($this->configuration);
         $this->configureBuilder($this->builder);
+        //build mus be executed
         $this->builder->build($this->sourceFile(), $this->targetFile());
     }
 
+    /**
+     * We expect no warnings to occur in building. If you want to test
```
greg0ire: "in building" reads like inside the physical place where you work. I'd use either "during the build" or "while building the documentation"

tests/Parser/DocumentParserTest.php
```diff
@@ -42,7 +45,11 @@ public function testErrorWhenDirectiveThrowsException(): void
             ->method('getName')
             ->willReturn('code-block-name');
 
-        $environment->expects(self::once())
+        $environment->expects(self::any())
```
greg0ire: I think `->expects(self::any())` achieves nothing. Let's drop it and call `method()` directly?

```suggestion
        $environment
```

tests/BaseBuilderTest.php
linawolf:
```suggestion
     * We expect no warnings to occur during the build. If you want to test
```

tests/BaseBuilderTest.php
```diff
@@ -5,29 +5,61 @@
...
+     * We expect no warnings to occur during the build. If you want to test
+     * expected warnings to be found, override this method.
+     */
+    protected function configureExpectedWarnings(): void
+    {
+        $this->errorManager->expects(self::never())->method('warning');
+    }
+
+    /**
+     * We expect no errors to occur in building. If you want to test
```
greg0ire: Same here

linawolf:
```suggestion
     * We expect no errors to occur during the build. If you want to test
```

greg0ire (dismissed):

SenseException (dismissed):
:+1: 

Please rebase your branch.

linawolf:
@greg0ire @SenseException the rebasing was non-trivial. please have another look

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @linawolf !

PR #239, opened 2023-01-09
[FEATURE] Allow multiple whitespaces as separator in directives and links

linawolf:
In the TYPO3 docs we like to use 2 white spaces after the two dots `..  ` marking the start of a directive or link. This is achieved in the editors by pressing tab after the two dots. This way directive and their blocks are aligned when using 4 spaces for indentation:

```
..  my-directive:: something
    My directives content
```

Currently such a directive or link is not detected to be a directive. However RST Syntax allows this: https://docutils.sourceforge.io/0.4/docs/ref/rst/restructuredtext.html#whitespace

For the time being tabs are not possible as blocks indented with tabs are not recognized. Tabs are allowed but not recommended by the rst syntax. Shall we disallow tabs and maybe issue a warning when a file contains tabs?

### DIFF
--- a/lib/Parser/LineChecker.php
+++ b/lib/Parser/LineChecker.php
@@ -114,12 +114,12 @@ class LineChecker
 
     public function isComment(string $line): bool
     {
-        return preg_match('/^\.\.(?: [^_]((?:(?!::).)*))?$/mUsi', $line) > 0;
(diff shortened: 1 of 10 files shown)


greg0ire:
> Shall we disallow tabs and maybe issue a warning when a file contains tabs?

I think so yes, it's a better UX than letting the user hope this is actually going to do something.

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @linawolf !

PR #245, opened 2023-01-15
[TASK] Delegate escaping to Twig where possible

linawolf:
After this discussion in the Symfony documentation channel: https://symfony-devs.slack.com/archives/C3ASP9LJJ/p1673725412479199

Added tests that XSS is not possible. Escaping of output should be done as late as possible, usually in the twig template. This way it is always done in the same place and not forgotten. When no twig is used we have to do the escaping right before the HTML is combined.

### DIFF
--- a/lib/TextRoles/WrapperTextRole.php
+++ b/lib/TextRoles/WrapperTextRole.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\RST\TextRoles;
 
+use function htmlspecialchars;
 use function sprintf;
 
 class WrapperTextRole extends TextRole
@@ -29,6 +30,6 @@ class WrapperTextRole extends TextRole
 
     public function process(string $text): string
     {
-        return sprintf($this->wrap, $text);
+        return sprintf($this->wrap, htmlspecialchars($text));
     }
 }
--- a/tests/Functional/tests/render/code-textrole-no-xss/code-textrole-no-xss.rst
+++ b/tests/Functional/tests/render/code-textrole-no-xss/code-textrole-no-xss.rst
@@ -1,0 +1,5 @@
+``<script>alert('hello!')</script>``
+
+`<script>alert('hello!')</script>`
+
+:code:`$this->execute(){}`
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         "symfony/polyfill-mbstring": "^1.0",
         "symfony/string": "^5.3 || ^6.0",
         "symfony/translation-contracts": "^1.1 || ^2.0",
-        "twig/twig": "^2.15.3 || ^3.4.3"
+        "twig/twig": "^3.5.0"
     },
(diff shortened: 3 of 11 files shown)


tests/Functional/tests/render/code-textrole-no-xss/code-textrole-no-xss.rst
```diff
@@ -0,0 +1,5 @@
+``<script>alert('hello!')</script>``
+
+`<script>alert('hello!')</script>`
+
+:code:`$this->execute(){}`
```
linawolf, line 5: Right now
 ```
:code:`<script>alert('hello!')</script>`
``` 
fails due to the last part being interpreted as URL target. I will provide a follow up for this

Comment (linawolf) [reply tests/Functional/tests/render/code-textrole-no-xss/code-textrole-no-xss.rst:5]: This problem is unrelated and I did not want to mix it in here

**Comment by greg0ire** [comment]
Can you squash your commits together in one, and use a more accurate title? From reading it, I don't expect any changes in `lib`. From what I understand when reading the diff, a good title might be "Delegate escaping to Twig". The body of the commit message should explain why you're doing that.

**Comment by linawolf** [comment]
Yeah you are right. I couldn't get the tests green as it seems that htmlspecialchars is behaving differently in different PHP versions

File: lib/TextRoles/WrapperTextRole.php [hunk 2]
```diff
@@ -29,6 +30,6 @@ public function getName(): string
 
     public function process(string $text): string
     {
-        return sprintf($this->wrap, $text);
+        return sprintf($this->wrap, htmlspecialchars($text));
```
greg0ire, line 33: Sorry for my ignorance, from your commit message I understand that:
1. When `SpanProcessor` is used, Twig is used.
2. When `WrapperTextRole` is used, Twig is not used.

It's just a deduction though, and I'd like to understand a bit more why that's the case.

linawolf: The WrapperTextRole is currently used in a way that you wrap the content with an html tag. Up until now there was no escaping whatsover, so the danger of XSS.

There is currently no twig template involved in the wrapper textrole. I have however considered it might be better to use a twig template in such situations as well. This would be a follow up then

linawolf: See for example [tests/Functional/tests/render/code-role/code-role.html](https://github.com/doctrine/rst-parser/pull/245/files#diff-9178f76996b18b66b29363fa197c405d78999da36523d0929a540d61bb52986c) where you can see that output was not escaped until now

greg0ire (approved):

greg0ire:
@javiereguiluz please review

javiereguiluz:
I don't know about the internals ... but looking at the tests it looks like the change is correct 👍

stof (approved):

greg0ire:
Thanks @linawolf !

PR #246, opened 2023-01-19
[TASK] Remove dependency on gajus/dindent

linawolf:
The project has many bugs that influence testing, Especially https://github.com/gajus/dindent/issues/25 is causing problems.

The project does not seem well maintained.

Switch to manual whitespace removal + DOM comparing in functional test.

Deprecate option to indent files. This was never really possible as gajus was only required in dev, option was offered for production however. Projects wishing to indent files nicer can do so in post-processing HTML, using the Events


javiereguiluz:
Nice proposal!

Indenting is also causing all kinds of issues in tests of symfony-tools/docs-builder. Happy to see this gone. We might need to ~take inspiration from~ copy your code to make our tests more resilient to to space changes.

Thanks

composer.json
```diff
@@ -35,14 +35,14 @@
...
-        "gajus/dindent": "^2.0.2",
         "phpstan/phpstan": "^1.9",
         "phpstan/phpstan-deprecation-rules": "^1.0",
         "phpstan/phpstan-phpunit": "^1.2",
         "phpstan/phpstan-strict-rules": "^1.4",
         "phpunit/phpunit": "^9.5",
         "symfony/css-selector": "4.4 || ^5.2 || ^6.0",
-        "symfony/dom-crawler": "4.4 || ^5.2 || ^6.0"
+        "symfony/dom-crawler": "4.4 || ^5.2 || ^6.0",
+        "ext-dom": "*"
```
greg0ire: This should be the first item of this list (that's why Composer lint is unhappy)

lib/Configuration.php
```diff
@@ -253,14 +250,15 @@ public function setIgnoreInvalidReferences(bool $ignoreInvalidReferences): void
         $this->ignoreInvalidReferences = $ignoreInvalidReferences;
     }
 
+    /** @deprecated The indenter is not maintained anymore and was removed. Indenting is not supported anymore */
     public function setIndentHTML(bool $indentHTML): void
```
greg0ire: It's OK to remove this entirely IMO: 0.6.0 is a new major release (that's how semver works when the version number starts with a zero)

composer.json
linawolf: do we have a tool to normalize the composer.json? In TYPO3 we use `ergebnis/composer-normalize` but it is not installed here

greg0ire: That's what I used to normalize the Composer manifest initially. We may add it as a dev dependency, I just thought it wouldn't be worth it.

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @linawolf !

PR #248, opened 2023-01-21
[TASK] Refactor Link Rendering

linawolf:
Delegate link rendering to a dedicated LinkRenderer in preparation for integrating references into text roles.

### DIFF
--- a/lib/HTML/Renderers/LinkRenderer.php
+++ b/lib/HTML/Renderers/LinkRenderer.php
@@ -1,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\RST\HTML\Renderers;
+
+use Doctrine\RST\Renderers\LinkRenderer as AbstractLinkRenderer;
+
+final class LinkRenderer extends AbstractLinkRenderer
+{
+    /** @param mixed[] $attributes */
+    public function renderUrl(?string $url, string $title, array $attributes = []): string
+    {
+        return $this->environment->getTemplateRenderer()->render('link.html.twig', [
+            'url' => $this->environment->generateUrl((string) $url),
+            'title' => $title,
+            'attributes' => $attributes,
+        ]);
+    }
+}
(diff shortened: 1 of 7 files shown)


lib/HTML/Renderers/LinkRenderer.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\RST\HTML\Renderers;
+
+use Doctrine\RST\Environment;
+
+final class LinkRenderer extends \Doctrine\RST\Renderers\LinkRenderer
```
greg0ire: I'm surprised phpcs is OK with this :thinking:

lib/HTML/Renderers/LinkRenderer.php
```diff
@@ -0,0 +1,27 @@
...
+{
+    public function __construct(Environment $environment)
+    {
+        parent::__construct($environment);
+    }
```
greg0ire: What am I missing here?

lib/HTML/Renderers/LinkRenderer.php
```diff
@@ -0,0 +1,27 @@
...
+
+    /** @param mixed[] $attributes */
+    public function renderUrl(?string $url, string $title, array $attributes = []): string
+    {
+        $url = (string) $url;
```
greg0ire: Let's inline this on line 22?

lib/LaTeX/Renderers/LinkRenderer.php
```diff
@@ -0,0 +1,39 @@
...
+
+use function is_string;
+use function substr;
+
+final class LinkRenderer extends AbstractLinkRenderer
+{
+    public function __construct(Environment $environment)
+    {
+        parent::__construct($environment);
+    }
```
greg0ire: To be removed?

lib/HTML/Renderers/LinkRenderer.php
linawolf: php cs was not ok with naming the latter LinkRendererInterface...

greg0ire: Oh right. How about aliasing it to `LinkRendererInterface` at import time?

greg0ire (approved):

greg0ire:
Thanks @linawolf !

PR #249, opened 2023-01-21
[FEATURE] Introduce Symfony Console Command

linawolf:
To make it easier to use this parser out-of-the-box.

### DIFF
--- a/bin/doctrine-rstparser.php
+++ b/bin/doctrine-rstparser.php
@@ -1,0 +1,14 @@
+<?php
+
+declare (strict_types=1);
+
+require __DIR__.'/../vendor/autoload.php';
+
+use Doctrine\RST\Command\RenderCommand;
+use Symfony\Component\Console\Application;
+
+$application = new Application();
+
+$application->add(new RenderCommand());
+
+$application->run();
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -16,6 +16,17 @@ You can install the Doctrine RST Parser with composer:
 Basic Usage
 -----------
 
+Then you can run the following script on the console:
+
+.. code-block:: console
+
+    $ vendor/bin/doctrine-rstparser render path/to/rst-source path/to/output
+
+With ``vendor/bin/doctrine-rstparser render -h`` you can see all possible options.
+
+Advanced Usage
+--------------
+
 Here is an example script that demonstrates how to use the RST Parser. Create a file named ``rst-test.php``
 in the root of your project and paste the following code:
 
--- a/lib/Command/RenderCommand.php
+++ b/lib/Command/RenderCommand.php
@@ -1,0 +1,85 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\RST\Command;
+
+use Doctrine\RST\Builder;
+use Doctrine\RST\Configuration;
+use Doctrine\RST\Formats\Format;
+use RuntimeException;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Helper\FormatterHelper;
+use Symfony\Component\Console\Input\InputArgument;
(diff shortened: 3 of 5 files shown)


bin/doctrine-rstparser.php
```diff
@@ -0,0 +1,14 @@
...
+require __DIR__.'/../vendor/autoload.php';
+
+use Doctrine\RST\Command\RenderCommand;
+use Symfony\Component\Console\Application;
+
+$application = new Application();
+
+$application->add(new RenderCommand());
+
+$application->run();
```
greg0ire: Please configure your IDE to always add a newline at end of file, this will make the red pictogram disappear.

lib/Command/RenderCommand.php
```diff
@@ -0,0 +1,85 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\RST\Command;
+
+use Doctrine\RST\Builder;
+use Doctrine\RST\Configuration;
+use Doctrine\RST\Formats\Format;
+use http\Exception\RuntimeException;
```
greg0ire: What's this? :thinking:

lib/Command/RenderCommand.php
```diff
@@ -0,0 +1,85 @@
...
+{
+    public function __construct()
+    {
+        parent::__construct('render');
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->setHelp('This command allows you to render your documentation...')
```
greg0ire, line 32:
```suggestion
            ->setHelp('This command allows you to render your documentation.')
```

lib/Command/RenderCommand.php
```diff
@@ -0,0 +1,85 @@
...
+            ->addOption(
+                'deleteCache',
+                'd',
+                InputOption::VALUE_NONE,
+                'Delete the cache before rendering?',
```
greg0ire, line 37: "clear cache" might sound more familiar, at least to Symfony users. Also, let's use kebab case for option, just like for most unix commands?




```suggestion
                'clear-cache',
                'c',
                InputOption::VALUE_NONE,
                'Whether to clear the cache before rendering',
```

lib/Command/RenderCommand.php
```diff
@@ -0,0 +1,85 @@
...
+            )
+            ->addOption(
+                'format',
+                'f',
+                InputOption::VALUE_OPTIONAL,
```
greg0ire: Why would the value be optional? `--format` alone does not have much meaning, right?

```suggestion
                InputOption::VALUE_REQUIRED,
```

docs/en/index.rst
```diff
@@ -16,6 +16,17 @@ You can install the Doctrine RST Parser with composer:
 Basic Usage
 -----------
 
+Then you can run the following script on the console:
+
+.. code-block:: console
+
+    $ vendor/bin/doctrine-rstparser render "my/path/to/documentation" "my/path/to/output"
```
greg0ire:
```suggestion
    $ vendor/bin/doctrine-rstparser render path/to/documentation path/to/output
```

greg0ire: I wonder if "documentation" is the right word here… I mean the output is also documentation, right? Maybe "source"?

lib/Command/RenderCommand.php
linawolf: No idea...

bin/doctrine-rstparser.php
linawolf: When I started using PHP I was told to NEVER, EVER put a new line at the end of a PHP file (as it was behind the closing php tag and caused an output of a new line). How times are changing... Configured

greg0ire (approved):

greg0ire:
Thanks @linawolf !

PR #251, opened 2023-01-22
[TASK] Refactor Configuration and Builder

linawolf:
- get rid of the Kernel class that was never much more then a factory for directives and text roles
- move configuration of index file name to Configuration
- Create directives and text roles with a collection of factories, part of them general factories, part format specific and also easily extendable just by adjusting the configuration


lib/Builder.php
```diff
@@ -23,11 +24,19 @@
 use function is_dir;
 use function sprintf;
 
+/**
+ * Builds a complete manual or book containing of multiple `.rst` documents
```
greg0ire:
```suggestion
 * Builds a complete manual or book containing multiple `.rst` documents
```

lib/Event/PostBuilderInitEvent.php
```diff
@@ -0,0 +1,39 @@
...
+
+use Doctrine\Common\EventArgs;
+use Doctrine\RST\Builder;
+use Doctrine\RST\Configuration;
+
+/**
+ * This event is called after the Builders initialization is finished,
+ * before the actual build starts.
+ *
+ * It replaces the former method Kernel::initBuilder()
```
greg0ire: This comment is useful for reviewers of this PR, but I think it should be removed from the source code, since `Kernel` itself is removed. If people outside the library need to know about this, it could go in `UPGRADE.md`, which is where it's fine to keep a bit of history.

Same for similar comments below.

greg0ire (approved):

linawolf:
@greg0ire I would be happy if we could get this merged as it is bound to cause lots of conflicts.

greg0ire:
Let's spare you the headache then I guess :slightly_smiling_face:

PR #252, opened 2023-01-27
[FEATURE] Introduce field lists

linawolf:
Field lists can be used standalone: https://docutils.sourceforge.io/docs/user/rst/quickref.html#field-lists or in the content area of a directive. They are not the same as the field options which belong directly to the directive.

### DIFF
--- a/lib/NodeFactory/NodeFactory.php
+++ b/lib/NodeFactory/NodeFactory.php
@@ -13,6 +13,7 @@ use Doctrine\RST\Nodes\ContentsNode;
 use Doctrine\RST\Nodes\DefinitionListNode;
 use Doctrine\RST\Nodes\DocumentNode;
 use Doctrine\RST\Nodes\DummyNode;
+use Doctrine\RST\Nodes\FieldListNode;
 use Doctrine\RST\Nodes\FigureNode;
 use Doctrine\RST\Nodes\ImageNode;
 use Doctrine\RST\Nodes\ListNode;
@@ -32,6 +33,7 @@ use Doctrine\RST\Nodes\TocNode;
 use Doctrine\RST\Nodes\WrapperNode;
 use Doctrine\RST\Parser;
 use Doctrine\RST\Parser\DefinitionList;
+use Doctrine\RST\Parser\FieldOption;
 use Doctrine\RST\Parser\LineChecker;
 use Doctrine\RST\Parser\ListItem;
 use Doctrine\RST\Parser\TableSeparatorLineConfig;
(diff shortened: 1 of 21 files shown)


lib/NodeFactory/NodeFactory.php
```diff
@@ -75,6 +77,9 @@ public function createSpanNode(Parser $parser, $span): SpanNode;
 
     public function createDefinitionListNode(DefinitionList $definitionList): DefinitionListNode;
 
+    /** @param FieldOption[] $items */
+    public function createFieldListNode(array $items): FieldListNode;
```
greg0ire, line 81: This should be documented in the UPGRADE.md I suppose, as it is a breaking change

lib/Parser/FieldOption.php
```diff
@@ -20,6 +24,15 @@ final class FieldOption
     /** @var string */
     private $body;
 
+    /**
+     * It depends on the context of wether the body of a field option should
```
greg0ire:
```suggestion
     * It depends on the context of whether the body of a field option should
```

lib/Parser/FieldOption.php
```diff
@@ -20,6 +24,15 @@ final class FieldOption
...
+     * be interpreted as nodes or plain text. Therefore, we store both and
+     * delicate the decision on which to use to the rendering.
```
greg0ire:
```suggestion
     * delegate the decision on which to use to the rendering.
```

lib/Parser/LineDataParser.php
```diff
@@ -217,12 +210,67 @@ public function parseDefinitionList(array $lines): DefinitionList
...
+            $fieldOption = new FieldOption($match['name'], strlen($match['offset']), $match['body'] ?? '');
+            $fieldOption->setNodes($this->getSubNodes($fieldOption->getBody()));
+
+            return $fieldOption;
         }
 
         return null;
     }
+
+    /** @return array|Node[] */
```
greg0ire: `Node[]` is an `array`, so this is as good as `array` :thinking:

tests/Renderer/FieldListNodeRendererTest.php
```diff
@@ -0,0 +1,34 @@
...
+use PHPUnit\Framework\TestCase;
+
+final class FieldListNodeRendererTest extends TestCase
+{
+    private FieldListNode $listNode;
+    private FieldListNodeRenderer $fieldListNodeRenderer;
+
+    protected function setUp(): void
+    {
+        $enviroment = new Environment(new Configuration());
```
greg0ire:
```suggestion
        $environment = new Environment(new Configuration());
```

lib/Parser/LineDataParser.php
linawolf: Oh yeah, my IDE is sometimes doing these strange automatic comments...

greg0ire (approved):

greg0ire:
Thanks @linawolf !

PR #257, opened 2023-02-04
[TASK] Refactor text role handling

### LINKED ISSUE #189: Document how to add "references"
Such as 

```
:issue:`12345` 
or
:php:`\My\Super\Duper\Class`
```

linawolf:
Unify handling of text roles with general syntax i.e. 
```
:emphasis`italic text`
```

and special syntax, i.e.:

```
*italic text*
```

Delegate rendering and processing to a specialized text role for each use case instead of having them all in the SpanProcessor  and SpanNodeRenderer. It is now possible to exchange a single text role with your own default text role.

Delegate link rendering to the text role rendering.

resolves #189


lib/TextRoles/BaseTextRole.php
```diff
@@ -47,4 +48,20 @@ public function setAliases(array $aliases): void
...
+    /*
+     * Does this text role have a special syntax like ``*cursive*``?
+     */
+    public function hasSpecialSyntax(): bool
+    {
+        return false;
+    }
+
+    /**
+     * @return array[string, array<SpanToken>]
```
greg0ire: I can use the code suggestion feature when reviewing commit by commit apparently, but this does not look like valid syntax. Should be `array{string, list<SpanToken>}` I think. There are more occurrences of this in the PR.

lib/TextRoles/SpecialTextRole.php
```diff
@@ -0,0 +1,42 @@
...
+     */
+    public function hasSpecialSyntax(): bool
+    {
+        return true;
+    }
+
+    public function replaceTokens(SpanProcessor $spanProcessor, string $span, string $pattern): string
+    {
+        $span = (string) preg_replace_callback(
+            $pattern,
```
greg0ire, line 30: Can `$pattern` be supplied by a user of this library? If yes, I'd expect `preg_last_error()` to be called at some point.

lib/TextRoles/BaseTextRole.php
linawolf: Mmh, the ci does not  complain, I have always used it like that and it always worked. Where do you have the other syntax from?

lib/TextRoles/TextRole.php
```diff
@@ -51,5 +51,10 @@ public function render(Environment $environment, SpanToken $spanToken): string;
      */
     public function hasSpecialSyntax(): bool;
 
-    public function getTokens(SpanProcessor $spanProcessor, string $span): string;
+    /**
+     * Replaces all findings of the special syntax of this text role with a unique string and registers the according
```
greg0ire: findings => occurrences

lib/TextRoles/SpecialTextRole.php
linawolf: I think it should be `protected` as only implementing text roles can use this

lib/Environment.php
```diff
@@ -145,6 +146,14 @@ public function getTextRole(string $section): ?TextRole
         return $this->textRoles[$section];
     }
 
+    /** @return TextRole[] */
+    public function getSpecialTextRoles(): array
+    {
+        return array_filter($this->textRoles, static function ($value) {
```
greg0ire: Please add type declarations

lib/TextRoles/LinkTextRole.php
```diff
@@ -158,4 +161,67 @@ private function registerLinkFromMatches(array $match, bool $anonymous = false):
...
         return $prev . $id . $next;
     }
+
+    private function replaceStandaloneHyperlinks(string $span): string
+    {
+        // Replace standalone hyperlinks using a modified version of @gruber's
+        // "Liberal Regex Pattern for all URLs", https://gist.github.com/gruber/249502
+        $absoluteUriPattern = '#(?i)\b((?:[a-z][\w\-+.]+:(?:/{1,3}|[a-z0-9%]))('
+            . '?:[^\s()<>]+|\(([^\s()<>]+|(\([^\s()<>]+\)))*\))+(?:\(([^\s()<>'
+            . ']+|(\([^\s()<>]+\)))*\)|[^\s\`!()\[\]{};:\'".,<>?«»“”‘’]))#';
```
greg0ire: Just a thought, but it could be cool to use the `x` modifier, that way we can use linebreaks inside the regex instead of doing this ugly concatenation (same below).

lib/TextRoles/TextRole.php
```diff
@@ -51,6 +51,12 @@ public function render(Environment $environment, SpanToken $spanToken): string;
      */
     public function hasSpecialSyntax(): bool;
 
+    /*
+     * Can the syntax also be applied to the text of some special text roles?
+     * For example withing link texts or table cells.
```
greg0ire: "within"

lib/TextRoles/LinkTextRole.php
linawolf: I did not introduce this regex, just move it. So I would suggest to make it beautiful in a follow up. I am not really experienced with regex so I doubt I can do it...

lib/Renderers/SpanNodeRenderer.php
```diff
@@ -55,7 +55,7 @@ private function renderTokens(string $span): string
     {
         $previousSpan = '';
         $i            = 0;
-        while ($previousSpan !== $span and $i < self::MAX_TOKEN_ITERATION) {
+        while ($previousSpan !== $span && $i < self::MAX_TOKEN_ITERATION) {
```
greg0ire, line 58: Please do an interactive rebase to fix each change in the relevant commit

lib/TextRoles/LinkTextRole.php
greg0ire: That's fine, I'll do it

lib/TextRoles/BaseTextRole.php
greg0ire: Psalm and PHPStan

greg0ire (approved):

greg0ire:
Thanks @linawolf !

lib/TextRoles/LinkTextRole.php
```diff
@@ -0,0 +1,274 @@
...
+
+                return $link;
+            }
+        }
+
+        return $this->renderLink($environment, $url, $linktext);
+    }
+
+    /** @param mixed[] $attributes */
+    abstract public function renderLink(Environment $environment, ?string $url, string $title, array $attributes = []): string;
```
HeahDude, line 65: How can this work with attributes if they are not passed line 61?

PR #258, opened 2023-02-04
Maintenance

greg0ire:


greg0ire:
The upgrade to PHPUnit 10 is difficult and should be done separately (it's difficult partly because we still support PHP 7.2)

linawolf (approved):

PR #261, opened 2023-02-04
Address deprecation of utf8_{de,en}code()

greg0ire:
These functions are deprecated as of PHP 8.2, because they don't force the user to explicitly state both the source and target encoding. Anyway, I think this was just a hack to make `substr()` work with multi-byte characters. Instead of using a hack, let us use `mb_substr()`.

Note that I did not switch the entire library to `mb_*()` functions because it broke some tests, but it should probably be done at some point.

See https://wiki.php.net/rfc/remove_utf8_decode_and_utf8_encode

### DIFF
--- a/lib/Nodes/TableNode.php
+++ b/lib/Nodes/TableNode.php
@@ -21,6 +21,7 @@ use function explode;
 use function implode;
 use function ksort;
 use function max;
+use function mb_substr;
 use function preg_match;
 use function sprintf;
 use function str_repeat;
@@ -28,7 +29,6 @@ use function strlen;
 use function strpos;
 use function substr;
 use function trim;
-use function utf8_decode;
 
 class TableNode extends Node
 {
@@ -133,7 +133,7 @@ class TableNode extends Node
             throw new LogicException('Cannot push data after TableNode is compiled');
         }
 
-        $this->rawDataLines[$this->currentLineNumber] = utf8_decode($line);
+        $this->rawDataLines[$this->currentLineNumber] = $line;
         $this->currentLineNumber++;
     }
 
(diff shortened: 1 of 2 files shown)


SenseException (approved):

lib/Nodes/TableNode.php
```diff
@@ -360,7 +360,7 @@ private function compilePrettyTable(): void
                 }
 
                 if ($currentColumnStart !== null) {
-                    $gapText = substr($line, $previousColumnEnd, $start - $previousColumnEnd);
+                    $gapText = mb_substr($line, $previousColumnEnd, $start - $previousColumnEnd);
```
stof, line 363: this looks wrong to me. `$start` and `$end` seem to be byte-based offsets (otherwise the check on line 356 is the one being wrong) while `mb_substr` works with codepoint-based offsets.

greg0ire: `$start` and `$end` are column number I think… their values are created by analyzing the separator line of a table, If I understand this properly: https://github.com/doctrine/rst-parser/blob/49513cbe07965277bc84716813b0f53255ed973f/lib/Parser/TableSeparatorLineConfig.php#L58-L72 

I think the check on line 356 might be the one being technically wrong: changing it to `mb_strlen()` does not break any test. I'm not sure what test to write to make it break though.

stof: As long as tests are working with ASCII only, they won't catch mistakes about mixing bytes vs codepoints (as ASCII codepoints are using a single byte in UTF-8)

greg0ire: The tests aren't ASCII only, see for instance https://github.com/doctrine/rst-parser/blob/0.5.x/tests/Functional/tests/render/table-grid-utf8/table-grid-utf8.rst

PR #266, opened 2023-03-09
Add support for fig* options in FigureNode

wouterj:
I was too fast, there is one other "bug" in 0.6: reStructuredText supports some options prefixed with `fig` on a figure node. These options apply to the wrapping `<figure>` element, rather than the ``.


greg0ire (approved):

linawolf (approved):

greg0ire:
Thanks @wouterj !

PR #267, opened 2023-03-12
Do not register references as link targets

wouterj:
The `` `...`_ `` syntax is a reference. The class was registering these as link targets in the meta. This resulted in many "duplicate link target" warnings when rendering the Symfony docs on 0.6 (as link targets cannot be duplicated across a project, but references can).


linawolf (approved):
Looks good to me

greg0ire:
Thanks @wouterj!

PR #273, opened 2023-08-01
'Cannot find column for index "0"' on TableNode rowspan

SenseException:
This is is referencing the issue of https://github.com/doctrine/doctrine-website/pull/476#issuecomment-1652288583 that is blocking the last steps of the PR doctrine/doctrine-website#476. It's also mentioned and described in https://github.com/doctrine/dbal/issues/3338#issuecomment-657255122.

The first commit contains rst and html files for a failing test.


SenseException:
@weaverryan I was looking through the code of the `TableNode` class and I don't have the impression that this table could be accomplished with the current  rowspan handling in `compilePrettyTable()`. Maybe you have an idea if this can be done with the current code because of the major changes you did a few years ago?

weaverryan:
@SenseException I don't have any idea, sorry - it's just been too long since I worked on that code. My rewrite was largely to try to take the old code, which was even more of a mess, and refactor it to be a bit more readable and meaningful. But 4 years later, I'm afraid I won't be much help

SenseException:
Thanks for taking a look at this. I currently see that another rewrite would be needed to get the expected rowspan for this use case :disappointed:

PR #280, opened 2024-01-31
Normalize paths to run on windows

srsbiz:
On windows all paths returned by build-in functions or SplFile object methods contain \ as directory separator. This breaks many things, like calculating relative directory for file being parsed, to include other files.

This can be easily fixed by using `Symfony\Component\Filesystem\Path::normalize` on any path that may be used to compare against other strings.

Windows path starts with partition letter followed by colon, so there is potential BC break in `FileIncluder.php`, where colon was used as path separator for `$includeRoot` argument. This feature to pass multiple roots was undocumented and there were no tests for it, so I hope this change will not have much impact.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,10 @@
     "autoload": {
         "psr-4": {
             "Doctrine\\RST\\": "lib/"
-        }
+        },
+        "files": [
+            "polyfill.php"
+        ]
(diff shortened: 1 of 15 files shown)


Oviglo:
Are there any news about this PR ?

PR #283, opened 2026-03-12
Bump dependencies

paulinevos:
To enable use with newer Symfony versions. Hopefully this will magically pass with no extra work.

Update: it magically passed with no extra work.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -27,6 +27,8 @@ env:
 jobs:
   phpunit:
     name: "PHPUnit"
-    uses: "doctrine/.github/.github/workflows/continuous-integration.yml@3.0.0"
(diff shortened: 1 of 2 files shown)


derrabus:
What about the Symfony components in `require-dev`?

greg0ire:
Not sure what's wrong with the PHPStan job :thinking:

SenseException:
> Not sure what's wrong with the PHPStan job 🤔

The PHPStan workflow isn't using our .github template. The current workflow is using ubuntu-20.04, which [support was stopped](https://github.com/actions/runner-images). I think we can switch to our standard workflow.

greg0ire (approved):

greg0ire:
Thanks @paulinevos !

greg0ire:
Released as 0.5.7

PR #289, opened 2026-03-30
Bump doctrine/coding-standard

greg0ire:
Fixes an issue I spotted on https://github.com/doctrine/rst-parser/pull/288

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         "twig/twig": "^2.9 || ^3.3"
     },
     "require-dev": {
-        "doctrine/coding-standard": "^11.0",
+        "doctrine/coding-standard": "^12.0 || ^14.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -34,7 +34,7 @@
         "twig/twig": "^2.9 || ^3.3"
     },
     "require-dev": {
-        "doctrine/coding-standard": "^11.0",
+        "doctrine/coding-standard": "^12.0 || ^14.0",
```
GromNaN, line 37: Why not restrict to the latest version?
```suggestion
        "doctrine/coding-standard": "^14.0",
```

greg0ire: Because then the package cannot be installed on PHP 7.2, which we still test against.

SenseException (approved):

GromNaN (approved):

PR #294, opened 2026-09-03
Run the end of life PHP jobs without Composer advisory blocking

### LINKED ISSUE #293: CI cannot resolve dependencies on PHP < 8.1 since Twig releases were flagged by security advisories
`composer update` fails on every PHP version below 8.1, so the four `PHPUnit (php, highest)` jobs for 7.2, 7.3, 7.4 and 8.0 cannot run. The last green CI run on `0.5.x` is from 30 March; the run of 13 June already failed this way, and every pull request opened since inherits a red check that has nothing to do with it.

## Cause

Composer blocks packages affected by a security advisory (`audit.block-insecure`, on by default). Every `twig/twig` release below 3.27 is flagged, and 3.27 requires PHP >= 8.1. The root constraint is `^2.9 || ^3.3`, so on PHP < 8.1 nothing is left:

```
Problem 1
  - Root composer.json requires twig/twig ^2.9 || ^3.3 -> satisfiable by twig/twig[v3.27.0, v3.27.1, v3.28.0].
  - twig/twig[v3.27.0, ..., v3.28.0] require php >=8.1.0 -> your php version (7.2.34) does not satisfy that requirement.
```

Forcing an older Twig shows what the first message hides:

```
$ composer update --dry-run --with "twig/twig:3.8.0"
  - Root composer.json requires twig/twig ^2.9 || ^3.3, found twig/twig[v2.9.0, ..., v3.28.0]
    but these were not loaded, because they are affected by security advisories (...)
```

Reproduced locally on `0.5.x` with `composer config platform.php`, Composer 2.9.5:

| platform php | `composer update` |
| --- | --- |
| 7.2.34 | fails |
| 7.4.33 | fails |
| 8.0.30 | fails |
| 8.1.31 | resolves |

## Possible fixes

1. Set `audit.block-insecure` to `false` for those jobs. Keeps the matrix and the supported range as they are; CI knowingly installs an old Twig on PHP versions that are themselves end of life.
2. Drop PHP < 8.1 from the matrix, and from `require.php`, which is a supported-versions decision.
3. Raise the constraint to `twig/twig ^3.27`, which amounts to the same thing through the back door.

I would go for the first one on `0.5.x` and leave the support range alone, but it is your call. Happy to send the PR for whichever you prefer.

Amoifr:
Fixes #293

`composer update` fails on PHP 7.2, 7.3, 7.4 and 8.0, so those four CI jobs cannot run. Every Twig release below 3.27 is flagged by a security advisory, 3.27 requires PHP >= 8.1, and Composer blocks advisory-affected packages by default.

Those four versions now get their own job, which runs `composer config audit.block-insecure false` before installing. The reusable workflow keeps 8.1 and up untouched, and nothing is committed to `composer.json`, so a normal install of this repository is still protected.

They need a job of their own because the reusable workflow has no hook to run a command before the install, and `composer-options` cannot carry it: neither `--no-audit` nor `COMPOSER_AUDIT_BLOCK_INSECURE=0` lifts the block, only the config setting does. Verified on platform PHP 7.2.34 with Composer 2.9.5, where `twig/twig` then resolves to 3.11.3 for `highest` and 2.9.0 for `lowest`. The new job carries its own `lowest` run on 7.2.

Dropping 7.2 to 8.0 from the reusable workflow moves its `lowest` job to 8.1, and there `phpunit/phpunit` resolved to 8.5, whose mock generator cannot handle the union types this test suite uses (`ReflectionUnionType::getName()`). The `require-dev` constraint becomes `^7.5 || ^9.5` so that install picks 9.6 instead. Checked at lowest on platform 8.1.34, where the suite passes with PHPUnit 9.6.33, and on platform 7.2.34, which still resolves PHPUnit 7.5.0 with Twig 2.9.0 for the end of life job.

Adding an input to `doctrine/.github` and calling the reusable workflow twice would do the same thing without the extra job, happy to send that instead.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -29,6 +29,51 @@ jobs:
     name: "PHPUnit"
     uses: "doctrine/.github/.github/workflows/continuous-integration.yml@15.1.0"
     with:
-      php-versions: '["7.2", "7.3", "7.4", "8.0", "8.1", "8.2", "8.3", "8.4", "8.5"]'
+      php-versions: '["8.1", "8.2", "8.3", "8.4", "8.5"]'
     secrets:
       CODECOV_TOKEN: "${{ secrets.CODECOV_TOKEN }}"
+
+  # Every twig/twig release below 3.27 is flagged by a security advisory, and 3.27 requires
+  # PHP >= 8.1, so the PHP versions that are themselves end of life have nothing left to install
+  # while Composer blocks affected packages. The reusable workflow has no hook to run a command
+  # before the install, so they get their own job rather than turning the blocking off for
+  # everyone. See #293.
+  phpunit-eol-php:
+    name: "PHPUnit (end of life PHP)"
+    runs-on: "ubuntu-22.04"
+
+    strategy:
+      matrix:
+        php-version:
+          - "7.2"
+          - "7.3"
+          - "7.4"
+          - "8.0"
+        dependencies:
+          - "highest"
+        include:
+          - php-version: "7.2"
+            dependencies: "lowest"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v6"
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          php-version: "${{ matrix.php-version }}"
+          coverage: "none"
+          ini-file: development
+
+      - name: "Accept the advisory-affected Twig releases these PHP versions are left with"
+        run: "composer config audit.block-insecure false"
+
+      - name: "Install dependencies with Composer"
+        uses: "ramsey/composer-install@v4"
+        with:
+          dependency-versions: "${{ matrix.dependencies }}"
+          composer-options: "--prefer-dist"
+
+      - name: "Run PHPUnit"
+        run: "vendor/bin/phpunit"


derrabus:
Wasn't the general idea that we only do this on EOL PHP version?

Amoifr:
You are right, and it is off `composer.json` now. The four end of life versions get their own job that turns the blocking off just for itself, and the reusable workflow keeps 8.1 and up as they were. The `lowest` run stays on 7.2, where the reusable workflow had put it.

They need a job of their own because the reusable workflow has no hook to run a command before the install, and `composer-options` cannot carry it: neither `--no-audit` nor `COMPOSER_AUDIT_BLOCK_INSECURE=0` lifts the block, only the config setting does. Checked on platform PHP 7.2.34 with Composer 2.9.5. If you would rather add an input to `doctrine/.github` and call the reusable workflow twice, say so and I will send that instead and drop this job.

One thing I did not know when I opened this: it is not only a CI problem. `twig/twig` is a runtime requirement, so a consumer on PHP < 8.1 is blocked the same way today.

```
$ composer update    # requiring doctrine/rst-parser:^0.5, platform php 7.4.33
  - doctrine/rst-parser[0.5.0, ..., 0.5.7] require twig/twig ^2.9 || ^3.3 -> satisfiable by twig/twig[v3.27.0, v3.27.1, v3.28.0].
  - twig/twig[v3.27.0, ..., v3.28.0] require php >=8.1.0
```

Nothing in this repository changes that for them. Only dropping PHP < 8.1 or raising the Twig constraint would, which is the supported versions call you set aside on the issue.

Amoifr:
Two follow-ups.

First, a correction to what I wrote above: the `lowest` run does not stay on 7.2. The end of life job carries its own 7.2 lowest run, but dropping those versions from the reusable workflow moves its `lowest` job to 8.1, and that is the job that was failing. There `phpunit/phpunit` resolved to 8.5, whose mock generator cannot handle the union types this suite uses (`ReflectionUnionType::getName()`).

So `require-dev` now asks for `^7.5 || ^9.5`, which makes that install pick 9.6 instead. Checked at lowest on platform 8.1.34, where the suite passes with PHPUnit 9.6.33, and on platform 7.2.34, which still resolves PHPUnit 7.5.0 with Twig 2.9.0 for the end of life job. It is a dev only constraint, but since you asked to keep `composer.json` out of the audit change, say the word if you would rather solve it in the workflow instead.

The five end of life jobs were green on the previous run, the failures next to them were fail-fast cancellations.

Amoifr:
Could a maintainer approve the workflow run on this one? It has never had a CI run, so the fix cannot be judged on anything but the diff, and that is a shame for a change whose entire purpose is the CI.

For a sense of what it unblocks: #291 and #292 each show ten red PHPUnit jobs, which look alarming but are a single failure. The PHP 7.2 job cannot install (`twig/twig[v3.27.0, ..., v3.28.0] require php >=8.1.0`), and `fail-fast: true` cancels the nine others, which report only `The operation was canceled.` Their `Coding Standards` and `PHPStan` checks are green. Every pull request opened here inherits that same wall of red.

The approach itself is the one you agreed to in #293: the blocking is turned off only inside the job for the four end of life versions, nothing lands in `composer.json`, and 8.1 and up keep using the reusable workflow untouched.

Thanks for maintaining this, and no hurry from my side.

composer.json
greg0ire, line 1: The commit message of 2402971f57b168dd22e29d430315549a904dd075 and the change don't seem to match.

Amoifr: You are right, the message said what I wanted the change to achieve rather than what it does. It drops the PHPUnit 8 range and raises the 9.x floor to 9.5, and neither was mentioned.

Reworded in fb7af7c, same one line diff:

> Drop the PHPUnit 8 range so the lowest install gets 9.5
>
> The end of life versions now have their own job, so the reusable workflow runs its lowest install on PHP 8.1. There composer resolved phpunit 8.5, whose mock generator calls `ReflectionUnionType::getName()` and cannot build doubles for the union types this suite uses.
>
> Dropping the 8.x range leaves 7.5, which PHP 8.1 rules out, and 9.5, so that install lands on PHPUnit 9.5.

Thanks for approving the workflow run, by the way. It is the first one this branch ever got, and the five end of life jobs pass, along with `PHPUnit (8.1, lowest)`, which is the one this commit is about.

greg0ire: Can the first commit pass the CI build without the second one?

Amoifr: No, it cannot, and the second commit exists only because of the first.

The first commit takes 7.2 to 8.0 out of the reusable workflow, so its `lowest` install lands on PHP 8.1. There the old constraint `^7.5 || ^8.0 || ^9.0` resolves PHPUnit 8.5, whose mock generator cannot build doubles for the union types this suite uses. That is the `PHPUnit (8.1, lowest)` failure I described earlier in this pull request.

I checked it on the first commit alone rather than from memory: with `platform.php` set to `8.1.0` and `--prefer-lowest`, composer locks `phpunit/phpunit 8.5.12`.

Reordering does not help either, since the base is already red for the reason this pull request fixes.

So: happy to squash the two into one commit, since the second is a consequence of the first rather than an independent change. Or to keep them apart if you would rather read the workflow change and the constraint change separately. Your call, I will do whichever you prefer.

greg0ire: I prefer atomic commits.

Amoifr: Squashed into one commit, since the first one leaves the build red on its own.

